Issue #49, opened 2020-12-16
CreateJwtPayloadDictionary: Expecting parsing custom claim by claim valueType


thohng:
**Which exact version are you using?**
5.0.0-preview.2.13

**Describe the bug**

Currently implementation parsing value object by claimName (claim.Type)
https://github.com/DuendeSoftware/IdentityServer/blob/main/src/IdentityServer/Extensions/TokenExtensions.cs#L132-L155

leastprivilege:
merged. thanks!

PR #50, opened 2020-12-16
Parsing claim object by claim valueType

thohng:
To fix issue #49


leastprivilege:
Good catch ;) thanks!

PR #55, opened 2020-12-21
Add abstraction for retrieving the issuer name for the current request

leastprivilege:
* Adds the `IIssuerNameService`
* Sets the issuer name for the current request on the `ValidatedRequest` object

This will allow removing direct dependencies on `IHttpContextAccessor` in various places


brockallen:
Would there be any reason to make the `GetCurrent` async? I don't really want to, but I'm trying to anticipate someone asking.

leastprivilege:
yea..

leastprivilege:
@brockallen  There are various places where we could switch to the new service, but they would be all breaking changes. Let's discuss.

leastprivilege:
notes:

* Sometimes we call multiple extension methods, e.g. GetIdentityServerIssuerUri and GetIdentityServerBaseUrl (discovery endpoint) - is it worth grouping them together into a broader scope interface?

Issue #59, opened 2020-12-27
KeySerializer still depends on NewtonSoft


leastprivilege:
have a look if we can get rid of it

leastprivilege:
https://github.com/DuendeSoftware/IdentityServer/blob/c308d8303d35131e59d705ecd1eb4ec49d131691/src/Storage/Stores/Serialization/CustomContractResolver.cs#L15

https://github.com/DuendeSoftware/IdentityServer/blob/b956af9daae5dce79ee44ac5c207345ca6abee5e/src/IdentityServer/Services/Default/KeyManagement/KeySerializer.cs#L14

PR #61, opened 2020-12-28
Key Management updates

brockallen:
* remove use of json.net #59
* add check to prevent mis-configuration of EC keys with X509 containers


leastprivilege:
I think you started with the wrong branch..

brockallen:
odd. i blame sourcetree.

Issue #56, opened 2020-12-22
Reduce places where we use the IHttpContextAccessor


leastprivilege:
e.g. to call HttpContext.GetIdentityServerIssuerUri

leastprivilege:
part of #31

brockallen:
The majority of uses is for a set of extension methods like "GetIdentityServerBaseUrl" in the HttpContextExtensions class. These would needs to be replaced by services in DI. So if that's possible (still TBD), then I see this as a 2-phase process. Phase 1 we add those services and move the code from the HttpContextExtensions into those implementations, then reimplement the extenstions in terms of the new services, and mark them as [Obsolete] (since they're public and likely used), and then in some phase 2 (v7?) we remove the extensions and replace the call locations to use the new services. Something like that.

brockallen:
A PR was started to work on this and the first major area was the use of a bunch of http context extension methods related to per-request pathing and host names. Unfortunately there is still quite a bit of http context accessor usage throughout the code base for things like cookie access, client certificate, and other valid per-request values. I'm not sure if introducing similar request abstractions would help or not for those. We will have to investigate each one individually, I think.

PR #70, opened 2021-01-05
Ref/usage of http context accessor

stefannikolei:
#56


brockallen:
Great, thanks for this!

PR #71, opened 2021-01-06
spanificate ParseScopeString

stefannikolei:
Here i refactored the ParseScopeString Method

Here are the benchmark results:

|                       Method |               scopes |     Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------------- |--------------------- |---------:|---------:|---------:|-------:|------:|------:|----------:|
|            ParseScopesString |    sc(...)e1    [34] | 757.5 ns | 15.17 ns | 17.47 ns | 0.1926 |     - |     - |     808 B |
| ParseScopesWithoutTrimString |    sc(...)e1    [34] | 733.1 ns |  3.00 ns |  2.35 ns | 0.1850 |     - |     - |     776 B |
|                     NextSpan |    sc(...)e1    [34] | 463.1 ns |  1.97 ns |  1.84 ns | 0.0496 |     - |     - |     208 B |
|               NextSpanNoSort |    sc(...)e1    [34] | 157.5 ns |  0.60 ns |  0.53 ns | 0.0496 |     - |     - |     208 B |

The last could be another optimization. I didn't want to do it because i had no clue if the Sort is needed.

What do you think about those changes? There a probably many more possible optimizations in the hotpath


stefannikolei:
test failure seems unrelated...

josephdecock:
@stefannikolei, thanks for this contribution! I'd be really interested to get an overall picture of where we could most benefit from optimizations. What made you want to optimize this function in particular?

I haven't measured anything, but my intuition is that access to data stores is typically going to be the most costly part of most request processing that identity server performs. I would guess that in many situations, the nanoseconds that are saved by optimizations like this won't really be noticeable, because the data access will completely dominate. So again, what's the rationale for optimization like this? I'm also wondering what kind of results we'd get if we ran these benchmarks with .NET 7.

stefannikolei:
No clue. to long ago. Ill close it

PR #72, opened 2021-01-06
Refactor DefaultTokenCreationService

stefannikolei:
* Use ValueTask where appropriate
* spawn Tasks and await later where appropriate
* create a static instance of JsonWebTokenHandler. It does not save any instance data and thus does not need to be created every single time


leastprivilege:
Thanks! We will re-visit when we drop .NET Core 3.1 - i am not a huge fan of ifdefs...

stefannikolei:
If you want I can drop the ifdefs to get the rest in?

leastprivilege:
We switch to target .NET 6 - feel free to update...

stefannikolei:
@leastprivilege i updated it

stefannikolei:
@leastprivilege Did you have a chance to have a look at it. Is there anything I need to tweak?

brockallen:
Thanks for this, but we're not in a position to accept this PR. Thanks.

PR #110, opened 2021-02-16
Added the option for returning the discovery document and JWKS indented

ishchatul:
Added the option for returning the discovery document and JWKS json results indented for easy reading 


leastprivilege:
Hi, 

thanks for this!

We think using a browser plugin to pretty-print your JSON is the better option.

PR #139, opened 2021-03-04
Update Quickstart UI to use Razor Pages

leastprivilege:


leastprivilege:
Some open questions 

* do we need [ValidateAntiForgeryToken] ?
* what about the [SecurityHeaders] filter
* better way to generate links? (is the /index needed?)
* how to do the redirect for native apps plumbing?

brockallen:
> do we need [ValidateAntiForgeryToken] ?

no

> what about the [SecurityHeaders] filter

added and works fine (with small change for PageResult)

> better way to generate links? (is the /index needed?)

yes, "/Login/Index" for example is needed.

> how to do the redirect for native apps plumbing?

Only way I could see was to add one more redirect to the "redirect" page (Razor Pages FTL on this front)

PR #188, opened 2021-04-13
Support for dynamic OIDC authentication schemes

brockallen:
Normally authentication handlers for external providers are added into your IdentityServer using AddAuthentication() and AddOpenIdConnect(). This is fine for a handful of schemes, but the authentication handler architecture in ASP.NET Core was not designed for dozens or more statically registered in the DI system. At some point you will incur a performance penalty for having too many. 

This feature is designed to allow OpenID Connect providers to be loaded and configured dynamically from a store. This also addresses the need to re-compile and re-run your startup code that would otherwise have the providers statically registered.

If you are using our EntityFramework package for the configuration store, this will require a DB schema change.

Fixes: #77


maulik-modi:
@leastprivilege and @brockallen , Is this same as https://www.identityserver.com/products/dynamic-authentication-providers?

leastprivilege:
It's a built-in and improved version, yes. Will be part of the Enterprise Edition.

PR #193, opened 2021-04-14
Prefer explicitly registered signing keys over automatic key management keys

brockallen:
This PR changes the behavior for which signing key is chosen. With this change if an explicit signing key is registered, then it should be preferred over the keys from the automatic key management feature.

Fixes: #177


leastprivilege:
add note about migration to docs and release notes

brockallen:
docs PR submitted

PR #197, opened 2021-04-16
add support for user tenant check in authorize interaction response generator against acr tenant value

brockallen:
In the authorize interaction response generator we have special handling for the `idp` claim in the user's session and comparing that against our special `idp:tenant` processing in the acr_values request param, and if it does not match we require the user to login.

This PR adds a similar check (and thus more first class support) for our acr_values `tenant:tenantid` processing. In short, it checks for a `tenant` claim and compares to the acr value and if they don't match it forces the user to re-authenticate.


brockallen:
Add feature flag, and disable by default (since this might have a breaking change on someone).

brockallen:
Feature flag added, and docs updated.

Issue #173, opened 2021-03-31
Formal support for UI to run on different host


brockallen:

brockallen:
SPA login sample built. Need one small fix in IdentityServer pertaining to return URL validation.

PR #209, opened 2021-04-21
Allow full host name to be included in OidcReturnUrlParser's IsValidReturnUrl

brockallen:
When validating return URLs we only allowed local paths to be accepted. But there is desire to allow the UI for login to outside the IdentityServer host, thus when they receive the returnUrl on the login page the full URL will be passed (including the host). 

This PR now allows the exact IdentityServer host to be included in the returnUrl for validation.

Fixes: #173


brockallen:
Open question: Do we want this behind a feature flag? // @leastprivilege

brockallen:
Feature flag added on `UserInteractionOptions` and documented (in docs repo).

```
        /// <summary>
        /// Flag that allows return URL validation to accept full URL that includes the IdentityServer scheme and host name. Defaults to false.
        /// </summary>
        public bool AllowHostInReturnUrl { get; set; }

```

PR #263, opened 2021-06-05
Update Quickstart UI to use Razor Pages (take 2)

brockallen:
Updated PR from #139


brockallen:
last bit of cleanup is now done

PR #310, opened 2021-07-07
Allow token_endpoint from discovery document as audience for Jwt Bearer Client Authentication

Zortaniac:

The problem occurs if the Jwt Bearer Client Authentication with the default PrivateKeyJwtSecretValidator in combination with an IdentityServer setup that has different domains for the issuer and the (token) endpoints.

The discovery document generator allows this configuring a different issuer, but the PrivateKeyJwtSecretValidator is constructing the token endpoint for the audience check path by its own, by combining the issuer with the token endpoint path.

This results in the affect that the client can not use the token_endpoint from the discovery document as the audience for the jwt client authentication, since the PrivateKeyJwtSecretValidator is constructing a difference audience during the check.

**To Reproduce**

Setup an IdentityServer where the issuer is configured to a different host and jwt bearer client authentication:
```
var builder = services.AddIdentityServer(options =>
{
    options.IssuerUri = "https://different.host";
})
.AddJwtBearerClientAuthentication();
```

This will result in discovery document:
```
{
   "issuer":"https://different.host",
   "jwks_uri":"https://localhost:5001/.well-known/openid-configuration/jwks",
   "authorization_endpoint":"https://localhost:5001/connect/authorize",
   "token_endpoint":"https://localhost:5001/connect/token",
   ...
}
```

Do a token request with the [JwtBasedClientAuthentication sample](https://github.com/DuendeSoftware/Samples/tree/main/IdentityServer/v5/Basics/JwtBasedClientAuthentication).

Notice that the audience of the request is "https://localhost:5001/connect/token" while the IdentityServer is checking for "https://different.host/connect/token".

From a client perspective this is only fixable by hard coding the right URI.

To prevent any compatibility issues it would be wise to add an additional allowed audience as proposed in this PR.

### DIFF
--- a/src/IdentityServer/Validation/Default/PrivateKeyJwtSecretValidator.cs
+++ b/src/IdentityServer/Validation/Default/PrivateKeyJwtSecretValidator.cs
@@ -10,6 +10,7 @@ using Duende.IdentityServer.Configuration;
 using Duende.IdentityServer.Models;
 using Duende.IdentityServer.Services;
 using Duende.IdentityServer.Extensions;
+using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
 using Microsoft.IdentityModel.JsonWebTokens;
 using Microsoft.IdentityModel.Tokens;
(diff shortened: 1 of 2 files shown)


leastprivilege:
We will look into it. Thanks!

brockallen:
Thanks for this and good catch. We need this PR against our patch release, so we opened a different PR for this. Will close this one. Thanks!

PR #328, opened 2021-07-13
Add cancellation token support to async methods

damianh:
Adds `CancellationToken` parameter with default value to async methods, pass token where it can, and update implementation & callsites as necessary.

This is a now-or-never type of change for V6 (.NET 6 LTS). Opening as draft PR for feedback before investing further time in it.

- [x] Storage interfaces
- [ ] Service interfaces

Also replaces some xml comments with `<inheritdoc/>` on some methods I touched where it was clearly a duplicated comment.

This will be a breaking change for interface implementers - they will need to include the CancellationToken parameter. Callsites would not be affected (compilation compatibility).


brockallen:
Update on this -- I submitted a PR to @damianh's branch to get a few more store interfaces. Once that's merged, this PR can get merged.

damianh:
Merged your PR.

brockallen:
So I've been looking at this more, and it turns out to do this in the way this PR is going, we may need to push this up thru every interface. IOW, there are places in various services where the CT is not being passed because the higher order services themselves don't accept a CT param. 

@damianh we previously discussed a different approach where the actual store implementation (e.g. the EF stores) would be responsible for obtaining a CT and just passing it to the EF APIs. They would obtain the CT via a service in DI, but the actual implementation would get it from the IHttpContextAccessor. And then we'd not have to change our 40+ interfaces. This pattern is in essence what ASP.NET Identity uses in the UserManager APIs (for better or for worse), so there's some bit of precedent. 

Thoughts? Of course, we can discuss this in more detail once everyone's back from holidays.

damianh:
Some quick on-vacation thoughts.

> So I've been looking at this more, and it turns out to do this in the way this PR is going, we may need to push this up thru every interface.
> IOW, there are places in various services where the CT is not being passed because the higher order services themselves don't accept a CT param.

Yes this change would viral to all interfaces with async methods.

> actual store implementation (e.g. the EF stores) would be responsible for obtaining a CT and just passing it to the EF APIs. They would obtain the CT via a service in DI, but the actual implementation would get it from the IHttpContextAccessor

A motivation for me for this PR was alternative storages, of which EF would be just one. There is an assumption with the proposed approach is that the developer has to ensure that activation is per-request and all implementations are hosted in a web app with access to an http-context (or some other CT source). Personally, I have have a strong preference for not having to define/restrict object lifecycle concerns on interface implementors. However I can see how this can be deemed to be a reasonable tradeoff against the impact of such a change... assuming it fully works for all scenarios of course.

brockallen:
> There is an assumption with the proposed approach is that the developer has to ensure that activation is per-request and all implementations are hosted in a web app with access to an http-context (or some other CT source). Personally, I have have a strong preference for not having to define/restrict object lifecycle concerns on interface implementors.

I might not be following this, but the use of these interfaces are within the context of IdentityServer request processing, so from the definition standpoint there is no use (or meaning) outside of that context. If you do, then you're projecting semantics on the interfaces that were never part of their design.

damianh:
Yes you are following it perfectly. A thought experiment - if you were to the interfaces for the first time today, would they contain cancellation token? Nonetheless, the ship has sailed.

brockallen:
Hey @damianh -- I think this style proposed [here](https://github.com/DuendeSoftware/IdentityServer/pull/405) is a less intrusive approach and proposed this way given the history and evolution of the code base.

brockallen:
Closing since the other PR for this topic was merged.

PR #384, opened 2021-08-30
Store tasks to reuse

Kahbazi:

Store tasks and reuse them in order to reduce allocations.


brockallen:
Hi. Thanks for this PR. We did some extensive testing recently, and this type of change is not very significant for a perf gain. We have plans to revisit this whole approach in a future release, so for now we'll leave it as-is. Thanks.

Issue #390, opened 2021-09-03
Generate specific error in the unlikely event of duplicate records


brockallen:
In the EF stores for the case insensitivity scenarios.

brockallen:
Looking more into this, this is unnecessary as all the keys are marked as unique constraints anyway. So there never will be duplicates, per-se. There are still collation issues for crupdate/upsert store semantics, but that's being addressed differently.

PR #391, opened 2021-09-03
Add better error logging for duplicate entries in EF

brockallen:
Fixes: #390

### DIFF
--- a/src/EntityFramework.Storage/Stores/ClientStore.cs
+++ b/src/EntityFramework.Storage/Stores/ClientStore.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Linq;
 using System.Threading.Tasks;
+using Duende.IdentityServer.EntityFramework.Entities;
 using Duende.IdentityServer.EntityFramework.Interfaces;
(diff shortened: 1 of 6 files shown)


src/EntityFramework.Storage/Stores/ClientStore.cs
```diff
@@ -50,23 +51,32 @@ public ClientStore(IConfigurationDbContext context, ILogger<ClientStore> logger)
...
+            Client entity;
+            try
+            {
+                entity = results.SingleOrDefault(x => x.ClientId == clientId);
+            }
+            catch (InvalidOperationException)
+            {
+                Logger.LogError("Duplicate Client entries found for clientId {clientId}", clientId);
+                return null;
+            }
```
Kahbazi, line 77: How about checking the `Length` property instead of checking with exception?
```suggestion
            if (results.Length == 0)
            {
                return null;            
            }
            
            if (results.Length > 1)
            {
                Logger.LogError("Duplicate Client entries found for clientId {clientId}", clientId);
                return null;            
            }
            
            var model = results[0].ToModel();
```

Issue #184, opened 2021-04-09
Add CancellationToken support for store interfaces


leastprivilege:

stefannikolei:
@leastprivilege I could have a look at this

stefannikolei:
Should i add overloads (and use Default Interface Implementation - pointing to the method without cancellationToken) or just add the CancellationTokens to the current methods. But this will be invasive.. 

What direction do you want me to go?

leastprivilege:
Thanks! Let's wait a little bit more until we start with the 6.0 work. Which will be soon.

stefannikolei:
Ok if you want you can ping me then.

leastprivilege:
Ah - you fixed it for Microsoft as well ;) nice!

https://github.com/dotnet/aspnetcore/pull/32736/files

stefannikolei:
@leastprivilege adding the cancellationtoken support to all store interfaces leads to a big change. Should i go for it. Or is that time wasted?

brockallen:
That work looks like it was recently started here: https://github.com/DuendeSoftware/IdentityServer/pull/328

damianh:
Sorry I missed this issue. Hope I didn't step on toes @stefannikolei .

PR #405, opened 2021-09-13
Add cancellation token support at boundaries

brockallen:
This PR adds a service to DI that presents the cancellation token from the HTTP context to the boundaries where they might be needed, such as outbound DB calls to EF. 

Fixes: https://github.com/DuendeSoftware/IdentityServer/issues/184


src/IdentityServer/Services/Default/DefaultCancellationTokenService.cs
```diff
@@ -0,0 +1,17 @@
+using Microsoft.AspNetCore.Http;
+using System.Threading;
+
+namespace Duende.IdentityServer.Services.Default
+{
+    class DefaultCancellationTokenService : ICancellationTokenService
```
damianh, line 6: 🤔 Perhaps `HttpContextCancellationTokenService` ? Bit more intention revealing?

brockallen:
@damianh Thanks for the feedback -- beyond that, any other input? As we discussed, it's a half-way solution to the problem given the code base history. Do you feel like it's sufficient for your purposes?

damianh:
Need to get a deeper look at it. Should have just started a review there, hit the "comment now" button. Will get at it shorty.

src/Storage/Services/ICancellationTokenService.cs
```diff
@@ -0,0 +1,19 @@
...
+
+
+using System.Threading;
+
+namespace Duende.IdentityServer.Services
+{
+    /// <summary>
+    /// Service to provide CancellationToken for async operations.
+    /// </summary>
+    public interface ICancellationTokenService
```
damianh, line 12: I'm thinking perhaps `ICancellationTokenProvider` (not a hill to die on.)

damianh (approved):
Still can't help the feeling that I'd prefer have it as a method parameter. However this works, no changes to suggest.

brockallen:
> Still can't help the feeling that I'd prefer have it as a method parameter. However this works, no changes to suggest.

Yep, and this is the compromise to not breaking almost every interface/service we have defined. If we had it to do over again, I'd have been 100% in agreement with your original proposal. If at some point in the future we have a "break the world" release, we'd do it then (but I don't if that will ever happen).

brockallen:
And thank you for the feedback and time on this!

Issue #298, opened 2021-06-21
EF Enhancement?


leastprivilege:
There are 8 child tables with FK to Client Table - https://id4withclients.readthedocs.io/en/latest/_images/ClientAppRelatedTables.png

By using ASSplitQuery, this results in multiple queries without left join
https://docs.microsoft.com/en-us/ef/core/querying/single-split-queries?WT.mc_id=DOP-MVP-5001942#split-queries-1

_Originally posted by @maulik-modi in https://github.com/DuendeSoftware/IdentityServer/discussions/293_

maulik-modi:
`2021-06-18 12:54:37.995 +00:00 [INF] Executed DbCommand (6ms) [Parameters=[@__clientId_0='?'], CommandType='"Text"', CommandTimeout='30']
SELECT c0."Id", c0."ClientId", c0."Type", c0."Value"
FROM "Clients" AS c
INNER JOIN "ClientClaims" AS c0 ON c."Id" = c0."ClientId"
WHERE c."ClientId" = @__clientId_0

2021-06-18 12:54:38.003 +00:00 [INF] Executed DbCommand (6ms) [Parameters=[@__clientId_0='?'], CommandType='"Text"', CommandTimeout='30']
SELECT c0."Id", c0."ClientId", c0."Created", c0."Description", c0."Expiration", c0."Type", c0."Value"
FROM "Clients" AS c
INNER JOIN "ClientSecrets" AS c0 ON c."Id" = c0."ClientId"
WHERE c."ClientId" = @__clientId_0

2021-06-18 12:54:38.009 +00:00 [INF] Executed DbCommand (6ms) [Parameters=[@__clientId_0='?'], CommandType='"Text"', CommandTimeout='30']
SELECT c0."Id", c0."ClientId", c0."Provider"
FROM "Clients" AS c
INNER JOIN "ClientIdPRestrictions" AS c0 ON c."Id" = c0."ClientId"
WHERE c."ClientId" = @__clientId_0

2021-06-18 12:54:38.018 +00:00 [INF] Executed DbCommand (8ms) [Parameters=[@__clientId_0='?'], CommandType='"Text"', CommandTimeout='30']
SELECT c0."Id", c0."ClientId", c0."PostLogoutRedirectUri"
FROM "Clients" AS c
INNER JOIN "ClientPostLogoutRedirectUris" AS c0 ON c."Id" = c0."ClientId"
WHERE c."ClientId" = @__clientId_0

[...]

PR #409, opened 2021-09-14
Suppress EF warning about split query behavior

brockallen:
Now with split queries as a feature in EF, if your query is not explicit on the behavior then EF emits a warning. This suppresses the warning in the DbContext itself.

This [issue](https://github.com/DuendeSoftware/IdentityServer/issues/298) is relevant.

### DIFF
--- a/hosts/EntityFramework/Startup.cs
+++ b/hosts/EntityFramework/Startup.cs
@@ -31,15 +31,12 @@ namespace IdentityServerHost
                 // this adds the config data from DB (clients, resources, CORS)
                 .AddConfigurationStore(options =>
                 {
-                    options.ConfigureDbContext = builder =>
-                        builder.UseSqlServer(connectionString)
-                            .ConfigureWarnings(w => w.Ignore(new EventId[] { RelationalEventId.MultipleCollectionIncludeWarning }));
+                    options.ConfigureDbContext = builder => builder.UseSqlServer(connectionString);
                 })
                 // this adds the operational data from DB (codes, tokens, consents)
                 .AddOperationalStore(options =>
                 {
-                    options.ConfigureDbContext = builder => builder.UseSqlServer(connectionString)
-                        .ConfigureWarnings(w => w.Ignore(new EventId[] { RelationalEventId.MultipleCollectionIncludeWarning }));
+                    options.ConfigureDbContext = builder => builder.UseSqlServer(connectionString);
 
                     // this enables automatic token cleanup. this is optional.
                     options.EnableTokenCleanup = true;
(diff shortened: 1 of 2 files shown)


LeaFrock:
If single query has better performance, why not fix the warnings with  `AsSingleQuery` instead of just ignoring them?

For example,
``` diff
        /// <summary>
        /// Finds a client by id
        /// </summary>
        /// <param name="clientId">The client id</param>
        /// <returns>
        /// The client
        /// </returns>
        public virtual async Task<Duende.IdentityServer.Models.Client> FindClientByIdAsync(string clientId)
        {
            var query = Context.Clients
                        .Where(x => x.ClientId == clientId)
                        .Include(x => x.AllowedCorsOrigins)
                        .Include(x => x.AllowedGrantTypes)
                        .Include(x => x.AllowedScopes)
                        .Include(x => x.Claims)
                        .Include(x => x.ClientSecrets)
                        .Include(x => x.IdentityProviderRestrictions)
                        .Include(x => x.PostLogoutRedirectUris)
                        .Include(x => x.Properties)
                        .Include(x => x.RedirectUris)
+                       .AsSingleQuery()
                        .AsNoTracking();
            
            var client = (await query.ToArrayAsync(CancellationTokenProvider.CancellationToken))
                .SingleOrDefault(x => x.ClientId == clientId);
            if (client == null) return null;

            var model = client.ToModel();

            Logger.LogDebug("{clientId} found in database: {clientIdFound}", clientId, model != null);

            return model;
```

LeaFrock:
For the consideration of compatible to 3.1,

``` csharp
#if NET5_0_OR_GREATER
                .AsSingleQuery()
#endif
```

brockallen:
> If single query has better performance, why not fix the warnings with AsSingleQuery instead of just ignoring them?

I explained a bit why here: https://github.com/DuendeSoftware/IdentityServer/issues/298#issuecomment-911756807

LeaFrock:
Excuse me, I'm not following what you mean...

> So in my quick and dirty local tests show that the split query is 2x slower than no split query. 

Single query equals to 'no split query'. So I wonder why you don't use `AsSingleQuery`(not default `AsSplitQuery` for EF Core 5+) with `Include` explicitly,  as your test says it gets better perf.
What this PR does is only ignoring the warning, but not fixing the source.

Official Doc:  https://docs.microsoft.com/en-us/ef/core/querying/single-split-queries

brockallen:
The other thread I said that a single query was faster, that's the default behavior, but if for some reason someone wants split query they can change the default when the provider is registered. This PR simply hides the warning for the default.

LeaFrock:
> but if for some reason someone wants split query they can change the default when the provider is registered.

Oh well, it makes sense. Thanks for your answer.

PR #421, opened 2021-09-21
Add feature to the default cache to prevent concurrent reloading of the cache

brockallen:
If concurrent threads get a cache miss, then it's possible that they all attempt to re-populate the cache. This PR adds a new GetOrAdd API to the ICache interface which allows the implementation to implement the concurrency management to prevent more than one request from reloading the cache at a time.


brockallen:
TODO: add timeout to lock

brockallen:
@leastprivilege added a concurrency lock timeout: https://github.com/DuendeSoftware/IdentityServer/pull/421/commits/ff534a4a518d9d0fccb8ed40d7ab9c48387745bc

PR #501, opened 2021-11-06
test Bullseye log grouping

adamralph:
**What issue does this PR address?**

I wanted to see how the new [log grouping](https://github.com/adamralph/bullseye/issues/640) in Bullseye 4.0.0 looks for your build.

PR #502, opened 2021-11-06
fix Bullseye and SimpleExec refs

adamralph:

The versions in the csproj are not used. They are controlled by the top level `Directory.Build.targets`.

This totally threw me when I was trying to run the build for https://github.com/DuendeSoftware/IdentityServer/pull/501


leastprivilege:
thanks

PR #550, opened 2021-11-30
Handle TaskCanceledException

brockallen:
Now that we're using the CancellationToken from the HttpContext, we might occasionally experience a non-error TaskCanceledException. This PR handles it so as to prevent the host's unhandled exception filter from running.

// @damianh

### DIFF
--- a/src/IdentityServer/Hosting/IdentityServerMiddleware.cs
+++ b/src/IdentityServer/Hosting/IdentityServerMiddleware.cs
@@ -92,6 +92,10 @@ namespace Duende.IdentityServer.Hosting
                     return;
                 }
             }
+            catch (TaskCanceledException ex)
+            {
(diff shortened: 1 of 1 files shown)


damianh:
I think this is a reasonable work around. Just wondering if there might be any unintentional side-effects.... nothing springs to mind. 

Unfortunate that EF team won't deal with until V7 (which means .NET 8 for LTS version).

brockallen:
Yea, if we don't do something like this then we will also be emitting the bogus log entry. And with this approach it's in one place, as opposed to every single place we pass the CancellationToken into an IO bound API. I couldn't think of anything better, so if you think it's ok, then we'll :shipit: 

TY!

brockallen:
Closing -- it feels this should be up to the host (not our middleware) to decide how to filter this.

PR #557, opened 2021-12-02
Prevent ASP.NET Core from caching OIDC provider options

brockallen:
* adds a non-caching (cache purging) decorator for the IIdentityProviderStore as the default

 Fixes: https://github.com/DuendeSoftware/IdentityServer/issues/528


brockallen:
@leastprivilege please have a look at the debug warning message that was added (as per our conversation)

PR #635, opened 2022-01-06
Update AutoMapper to v11

brockallen:


neozhu:
Can you consider upgrading directly to automapper version 12?

leastprivilege:
Can you consider removing AutoMapper altogether?

Issue #645, opened 2022-01-12
Db context pooling on ConfigurationStore results in an exception


msschl:
**Which version of Duende IdentityServer are you using?**
v6.0.0

**Which version of .NET are you using?**
.Net 6.0

**Describe the bug**

As described by the [Advanced Performance Topics Docs](https://docs.microsoft.com/en-us/ef/core/performance/advanced-performance-topics?tabs=with-di%2Cwith-constant#limitations) the `OnConfiguring` method cannot be used with context pooling.
However, the `ConfigurationDbContext` uses the `OnConfiguring` method resulting in a runtime exception being thrown.

```cs
namespace Duende.IdentityServer.EntityFramework.DbContexts;

public class ConfigurationDbContext<TContext> : DbContext, IConfigurationDbContext
    where TContext : DbContext, IConfigurationDbContext
{
    ...

    protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
    {
        base.OnConfiguring(optionsBuilder);
        optionsBuilder.ConfigureWarnings(w => w.Ignore(new EventId[] { RelationalEventId.MultipleCollectionIncludeWarning }));
    }
}
```

**To Reproduce**

Enable context pooling for the `ConfigurationStore`.

brockallen:
> As described by the Advanced Performance Topics Docs the OnConfiguring method cannot be used with context pooling.

It says "DbContext pooling has a few limitations on what can be done in the OnConfiguring method of the context.". I don't see where it says "the OnConfiguring method cannot be used with context pooling".

> resulting in a runtime exception being thrown.

Can you share the exception and call stack details, please?

msschl:
```logs
System.InvalidOperationException: 'OnConfiguring' cannot be used to modify DbContextOptions when DbContext pooling is enabled.
   at Microsoft.EntityFrameworkCore.DbContext.get_ContextServices()
   at Microsoft.EntityFrameworkCore.DbContext.get_Model()
   at Microsoft.EntityFrameworkCore.Internal.InternalDbSet`1.get_EntityType()
   at Microsoft.EntityFrameworkCore.Internal.InternalDbSet`1.CheckState()
   at Microsoft.EntityFrameworkCore.Internal.InternalDbSet`1.get_EntityQueryable()
   at Microsoft.EntityFrameworkCore.Internal.InternalDbSet`1.System.Linq.IQueryable.get_Provider()
[...]

PR #646, opened 2022-01-12
Fix OnConfiguring exception from EF when DbContext pooling is enabled

brockallen:
When EF DbContext pooling is enabled for our ConfigurationDbContext, the EF options builder can't be modified more than one in OnConfiguring. This PR prevents us from trying to modify the options more than once.

Fixes:  #645


brockallen:
@leastprivilege please review, and if good then merge, please, and we can release (at your discretion).

leastprivilege:
looks great!

PR #663, opened 2022-01-19
Use the built-in CSPRNG GetInt32

vcsjones:
.NET Core 3.0+ have this capability built-in, so use that instead of rolling it.

### DIFF
--- a/src/IdentityServer/Services/Default/NumericUserCodeGenerator.cs
+++ b/src/IdentityServer/Services/Default/NumericUserCodeGenerator.cs
@@ -36,27 +36,7 @@ public class NumericUserCodeGenerator : IUserCodeGenerator
     /// <returns></returns>
     public Task<string> GenerateAsync()
     {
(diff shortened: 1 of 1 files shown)


src/IdentityServer/Services/Default/NumericUserCodeGenerator.cs
```diff
@@ -36,27 +36,7 @@ public class NumericUserCodeGenerator : IUserCodeGenerator
     /// <returns></returns>
     public Task<string> GenerateAsync()
     {
-        var next = Next(100000000, 999999999);
+        var next = RandomNumberGenerator.GetInt32(100000000, 1000000000);
```
vcsjones, line 39: The one built in to .NET has an exclusive upper bound, while the previous one was inclusive, hence the upper bound of `1_000_000_000`.

brockallen:
Yo, it's @vcsjones -- howdy sir!

brockallen:
LGTM

Issue #694, opened 2022-01-27
Seeing IdentityServer request slowness after upgrading from v5.2.3 to v6


billpieper:
We upgraded IdentityServer versions and made no other infrastructure changes, but we are seeing much slower response times post upgrade.  

We are using a SQL Azure database for the IdentityServer data store and Azure Kubernetes Service (AKS) to host IdentityServer via container.

Response time data from Application Insights when using IdentityServer v5.2.3:
![image](https://user-images.githubusercontent.com/682286/151389824-46433727-76cc-4162-9da0-b62338b9a6ff.png)

Similar view after updating to IdentityServer v6:
![image](https://user-images.githubusercontent.com/682286/151389897-08b95c93-401d-451d-83b1-d8323b9f4fd5.png)

In Issue #668 it was asked in caching was being used.  We had NOT implemented a cache up to this point. -- Yesterday, I enabled the in memory cache  (via service.AddIdentityServer().AddInMemoryCaching() per https://docs.duendesoftware.com/identityserver/v6/data/configuration/#caching-configuration-data)

Enabling in memory cache has helped a bit -- but the performance is still quite a bit slower than what we were seeing with v5.2.3:
![image](https://user-images.githubusercontent.com/682286/151390847-ca66b641-68a1-42ac-9d09-f07e354dfc2e.png)

brockallen:
Hi @billpieper -- What do you use for your user auth library/database (e.g. ASP.NET Identity)? I'd like to see a comparison of something that uses your user auth library (like the login page) of before and after outside of an OIDC workflow. In other words, how does that compare without any IdentityServer code running.

billpieper:
In our UI (vue js spa application) we are currently using oidc-client v1.9.1.
For users/database -- we federate with Azure AD and all our user identities originate there.

hugoqribeiro:
I've seen the same yesterday in our staging environment with the configuration store cache disabled (because of the other error) in the connect/authorize endpoint.
I could get to a point in Azure Application Insights where it seemed that there was some kind of contention around the queries that load the client and validate the resources configuration.
I wasn't able yet to look into the Azure SQL logs because that depends on our infra team.
Yeah... DevOps is still a pain... :)

brockallen:
Finally, what DB are you using for the stores @billpieper? SqlServer? Postgres? Or maybe you're not and it's all just in-memory?

billpieper:
[...]

PR #705, opened 2022-02-01
EF fixes

brockallen:
This PR fixes two EF issues:

* Client Store query has been reverted to the v5.x style, which makes multiple round trips to avoid the matrix multiplication when an inner join is used for the various child tables. Customers were experiencing worse performance with the inner join.

Fixes: #694

* Quickstart Admin Pages for Client Creation: They were missing storing the client secret and the AllowOfflineAccess settings.


brockallen:
> Have you tried AsSplitQuery? It works fine. I've confirmed that in our override of the client store.

Doesn't that have the same effect as the way the code is written now?

brockallen:
> Doesn't that have the same effect as the way the code is written now?

I just looked at the difference (at least on SqlServer). The AsSplitQuery doesn't reset the connection across the child tables queries, whereas the older approach does. I guess that'd end up being more efficient?

PR #734, opened 2022-02-14
Replace home grown logic of ToSpaceSeparatedString with String.Join

leastprivilege:
Replaces our home grown logic with the built-in String.Join implementation which is more efficient


brockallen:
I added a couple of unit tests -- just look at them to see if they match expectations (mainly the odd one).

PR #791, opened 2022-03-09
Make AuthorizationRequest ctor as public to override implementation IReturnUrlParser

yartat:
Make `AuthorizationRequest` ctor as public to override implementation `IReturnUrlParser`

AuthorizationRequest class contains public constructor without parameters and internal constructor with initialization from validated request parameters. I want to override `OidcReturnUrlParser` to process request URI's without validation for local URL because IS is hosted in AWS kubernetes.

### DIFF
--- a/src/IdentityServer/Models/Messages/AuthorizationRequest.cs
+++ b/src/IdentityServer/Models/Messages/AuthorizationRequest.cs
@@ -2,6 +2,7 @@
 // See LICENSE in the project root for license information.
 
 
+using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Linq;
@@ -109,7 +110,6 @@ public class AuthorizationRequest
     /// </value>
     public IEnumerable<Claim> RequestObjectValues { get; }
 
-
     /// <summary>
     /// Initializes a new instance of the <see cref="AuthorizationRequest"/> class.
     /// </summary>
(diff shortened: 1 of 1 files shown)


SergiiKram:
Hi @leastprivilege, would you please take a look at this issue?

To give you some context here: we need to implement our own `IReturnUrlParser` for some reason.
So we need to return an `AuthorizationRequest` instance from the `IReturnUrlParser.ParseAsync(string returnUrl)` method.

Unfortunately, the `AuthorizationRequest` class's constructor `AuthorizationRequest(ValidatedAuthorizeRequest request)` is internal. Alternatively, if we try to use the default public constructor, we cannot set the `Parameters` and `RequestObjectValues` properties outside because they do not have public setters.

We can work around it using reflection 
```csharp
return Activator.CreateInstance(
                        typeof(AuthorizationRequest),
                        BindingFlags.NonPublic | BindingFlags.Instance,
                        null,
                        new object[] { result.ValidatedRequest },
                        null) as AuthorizationRequest;
```
but this does not sound right.
We better have a public/protected constructor or public property setters.

If we are doing something very wrong here and trying to use the library in a way it was not designed to, please let us know.

Thank you.

brockallen (approved):

brockallen:
Actually, this should be done against the releases/6.1.x patch branch if you want it included in the upcoming patch. Mind rebasing it?

brockallen:
Ah, I can do that for you. Done and thanks!

SergiiKram:
Thanks, @brockallen.

Issue #723, opened 2022-02-10
Problems migrating to IdSrv 6 with EF Core 6


galvesribeiro:
**Which version of Duende IdentityServer are you using?**
Moving from 5.2.2 to 6.0.4

**Which version of .NET are you using?**
.Net 6 (6.0.101)

**Describe the bug**
After upgrading all Duende.IdentityServer(.*) packages to 6.0.4 and try to run `dotnet ef migrations add ...`, it fail with the following exception: 

```
Using context 'PersistedGrantDbContext'.
System.InvalidOperationException: Unable to resolve service for type 'Duende.IdentityServer.EntityFramework.Options.OperationalStoreOptions'. This is often because no database provider has been configured for this DbContext. A provider can be configured by overriding the 'DbContext.OnConfiguring' method or by using 'AddDbContext' on the application service provider. If 'AddDbContext' is used, then also ensure that your DbContext type accepts a DbContextOptions<TContext> object in its constructor and passes it to the base constructor for DbContext.
```

The only change on the codebase I've noticed is that both `ConfigurationDbContext` and `PersistedGrantDbContext` ctor lost the second parameter (`ConfigurationStoreOptions` and `OperationalStoreOptions ` respectively). If I revert back to 5.2.2 the exception doesn't happen and it add the migration perfectly.

**Expected behavior**

The migration being added just like on 5.2.2.

**Log output/exception with stacktrace**

```
% dotnet ef migrations add InitialIdentityServerPersistedGrantDbMigration -c PersistedGrantDbContext -o Data/Migrations/PersistedGrantDb --verbose
Using project '<PROJECT_DIR>/Identity.csproj'.
Using startup project '<PROJECT_DIR>/Identity.csproj'.
Writing '<PROJECT_DIR>/obj/Identity.csproj.EntityFrameworkCore.targets'...
dotnet msbuild /target:GetEFProjectMetadata /property:EFProjectMetadataFile=/var/folders/zv/f_g8wmd128sbhbkc8j4pmc4w0000gn/T/tmpM6BB9h.tmp /verbosity:quiet /nologo <PROJECT_DIR>/Identity.csproj
Writing '<PROJECT_DIR>/obj/Identity.csproj.EntityFrameworkCore.targets'...
[...]

PR #801, opened 2022-03-11
[EF] add back overloaded ctor for accepting store options for DbContexts

brockallen:
In v6.0 in our `DbContext` classes we had removed the ctor param for the options class and changed it to expect the value from DI. In some scenarios (like when using the `IDesignTimeContextFactory`), this wasn't DI driven and so not easily available. This add an additional overloaded ctor that accepts the options parameter.

Related: https://github.com/DuendeSoftware/IdentityServer/issues/723


leastprivilege (approved):

brockallen:
This is reverted by: https://github.com/DuendeSoftware/IdentityServer/pull/893

PR #820, opened 2022-03-16
Server-side session management features

brockallen:
Given the server-side session feature, we can build on top a few more session management features. This includes:

* triggering back-channel logout notifications when the user's session expires
* revoking tokens when the user's session expires or logs out
* extending the server-side session expiration upon token use
* rejecting tokens unless the user's session is still active

Combining these features, this allows IdentityServer to monitor and track a user's session across all clients being used, and thus coordinating session expiration across all of them. More specifically, when a user is idle for longer than the session expiration at IdentityServer, then they can be logged out everywhere. But if the user remains active at one of them, then it will extend the session.

This PR adds these features.


brockallen:
@leastprivilege For the `DefaultRefreshTokenService` code to slide the server-side session, perhaps a decorator rather than change the core implementation?

leastprivilege:
ok. that's an idea

leastprivilege (approved):

PR #858, opened 2022-04-15
Remove setter from Properties on IdentityProvider

brockallen:
It's possible to overwrite the `Properties` dictionary, thus un-setting other properties on the `IdentityProvider` class:

```
new OidcProvider
            {
                Scheme = "google",
                DisplayName = "Google",
                Authority = "https://accounts.google.com",
                ClientId = "...",
                Scope = "openid profile email",
                Properties = new Dictionary<string, string>
                {
                    { "foo", "bar" }
                }
            }
```

This PR removes the setter to prevent this problem. The updated code from above would now look like this:

```
new OidcProvider
            {
                Scheme = "google",
                DisplayName = "Google",
                Authority = "https://accounts.google.com",
                ClientId = "...",
                Scope = "openid profile email",
                Properties = 
                {
                    { "foo", "bar" }
                }
            }
```

### DIFF
--- a/migrations/IdentityServerDb/SeedData.cs
+++ b/migrations/IdentityServerDb/SeedData.cs
@@ -107,7 +107,11 @@ public class SeedData
                 DisplayName = "Google",
                 Authority = "https://accounts.google.com",
                 ClientId = "998042782978-gkes3j509qj26omrh6orvrnu0klpflh6.apps.googleusercontent.com",
-                Scope = "openid profile email"
+                Scope = "openid profile email",
+                Properties = 
+                {
+                    { "foo", "bar" }
+                }
             }.ToEntity());
             context.SaveChanges();
         }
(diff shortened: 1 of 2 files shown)


arothberg4:
This will help a lot.  Thanks!

In the future, I'd love to see more "defensiveness" in this area because if these properties get modified by the developer, it's not obvious this ALSO impacts the critical properties which are populated via the standard setters (e.g. calling Properties.Clear()). It still presents itself (IMHO) as only impacting the "custom" properties explicitly added by the developer.  And issues in the orchestration of authentication with the auth server and/or idp are **incredibly** difficult to debug.

For the few classes I can think of with a similar property (e.g., IdentityModel.TokenRequest.**Parameters**), I can't think of one where that collection is impacted by the the standard setters - it's usually isolated to just what the dev explicitly sets on the collection property.

So maybe that can be in a future release, that property can be exposed as an IReadOnlyDictionary with the properties only populated via the constructor in a provided "setup" Action<IdentityProviderOptions>?

leastprivilege (approved):

PR #862, opened 2022-04-18
minor reorg of session mgmt service/store/model names, add OTel support

brockallen:
Fixes: #852


brockallen:
@leastprivilege this is the last thing for 6.1, I think

leastprivilege (approved):

PR #900, opened 2022-05-23
add more defensive check for server side session store in DI

brockallen:
When using our EF package, the [optional by design] server side session store is always added to DI, even though `AddServerSideSessions` is not called. As long as none of the server-side session coordination features are enabled, then this has no effect. But it's possible that this is added, server-side sessions are not being used, but one of the session coordination features is accidentally enabled, then it could cause unintended side effects for the processing of the client's request.

This PR adds a more defensive check to not allow any of the server side session store logic to be used if `AddServerSideSessions` is not called.

### DIFF
--- a/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/SessionManagement.cs
+++ b/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/SessionManagement.cs
@@ -75,7 +75,14 @@ public static class SessionManagementServiceCollectionExtensions
     public static IIdentityServerBuilder AddServerSideSessionStore<T>(this IIdentityServerBuilder builder)
         where T : class, IServerSideSessionStore
     {
-        builder.Services.AddTransient<IServerSideSessionStore, T>();
+        builder.Services.AddTransient<IServerSideSessionStore>(svcs =>
+        {
+            if (svcs.GetService<IServerSideSessionsMarker>() == null) return null;
+            return svcs.GetRequiredService<T>();
+        });
+
+        builder.Services.AddTransient<T>();
+
         return builder;
     }
 }


leastprivilege (approved):

richieto:
I faced this problem, not by accident, I wanted to enable _**CoordinateClientLifetimesWithUserSession**_ without having _Server-Side Sessions_ and I was getting a validation error from here:
https://github.com/DuendeSoftware/IdentityServer/blob/6a45df770da336604801bddca593a902f1ccb802/src/IdentityServer/Services/Default/DefaultSessionCoordinationService.cs#L208
This was preventing the user to complete the login process

Upgrading to Duende 6.2 solved it nicely. 👍

oliverw:
As a side note, the constructor of `DefaultSessionCoordinationService` relies on the DI Container to inject a null value for the last constructor argument `IServerSideSessionStore serverSideSessionStore = null`. This works with the .NET Built-In DI but crashes if the project is using Autofac for example because Autofac does not seem to support injecting null value into constructors even if a default value of null has been specified for argument in question.

brockallen:
Hi @oliverw -- understood. But due to Microsoft's decision to implement DI as standard in ASP.NET Core, they have in fact normalized this behavior as how the platform works, and we rely on it in some circumstances.

narbit:
Just ran into another situation when both, `AddServerSideSessions` and `CoordinateClientLifetimesWithUserSession`, are enabled. If client is configured for `CoordinateLifetimeWithUserSession` AND supports both, interactive and non-interactive (ropc or custom) flows that produce refresh tokens, the refresh grant for non-interactive grants is broken because session validation in `DefaultSessionCoordinationService` fails with `Due to missing/expired server-side session...`

brockallen:
@narbit -- We don't see a lot of resource owner use cases these days due to it being deprecated, but I think our base class is extensible so you can override the behavior. I've not looked into it, but I think you get the full context of the token request so you could check the grant type and override that behavior.

narbit:
@brockallen , I was mostly asking because of various custom grants we have. I can override the `DefaultSessionCoordinationService` but unfortunately it doesn't give me the full token context. It doesn't look like I can override `ServerSideSessionRefreshTokenService` which has the token context. This is v6.2.0. I also submitted a support request/question for this via email.

narbit:
Ended up overriding `ValidateSessionAsync` in `DefaultSessionCoordinationService` while retrieving RT from `IHttpContextAccessor` and then checking AT `amr` claim for the grant type. But it'd be easier if `ValidateRefreshTokenAsync` was virtual in `ServerSideSessionRefreshTokenService` or if `TokenValidationResult` included claims.

josephdecock:
Thanks for this feedback @narbit  - we will consider making that method virtual in a future release.

Issue #949, opened 2022-06-27
Consider using something custom rather than ISystemClock


brockallen:
Context: https://github.com/DuendeSoftware/Support/issues/101

brockallen:
Or this: https://github.com/dotnet/aspnetcore/issues/47472

brockallen:
Related: https://github.com/DuendeSoftware/IdentityServer/issues/1203

brockallen:
Re-enable TreatWarningsAsErrors in Directory.Build.props

PR #950, opened 2022-06-27
ensure we capture milliseconds of the current DateTime when creating events

brockallen:
The Microsoft provided ISystemClock doesn't capture milliseconds. This PR will use DateTime directly to ensure events have milliseconds in the created/timestamp field.

Context: https://github.com/DuendeSoftware/Support/issues/101#issuecomment-1167779325

A larger effort to rework is considered for the next major release: https://github.com/DuendeSoftware/IdentityServer/issues/949


leastprivilege (commented):
DateTime or DateTimeOffset?

BrianAlk (approved):

leastprivilege (approved):

PR #961, opened 2022-07-02
Allow for naming tables by convention.

profet23:

Fixes https://github.com/DuendeSoftware/Support/issues/113

This PR will allow for naming EFCore tables via convention.

With this PR, this can be achieved with adding the following code to your `AddConfigurationStore()` and `AddOperationalStore()` methods.

```csharp
// Remove the IdentityServer default table names so that naming by convention (snake case) will be used.
foreach (var tableConfigurationProperty in options.GetType().GetProperties().Where(p => p.PropertyType == typeof(TableConfiguration)))
{
        if (tableConfigurationProperty.GetValue(options) is TableConfiguration tableConfiguration)
        {
	        tableConfiguration.Name = string.Empty;
        }
}
```

With the above code in place, you can use https://github.com/efcore/EFCore.NamingConventions as expected.

### DIFF
--- a/src/EntityFramework.Storage/Extensions/ModelBuilderExtensions.cs
+++ b/src/EntityFramework.Storage/Extensions/ModelBuilderExtensions.cs
@@ -37,7 +37,6 @@ public static class ModelBuilderExtensions
         modelBuilder.Entity<Client>(client =>
         {
             client.ToTable(storeOptions.Client);
-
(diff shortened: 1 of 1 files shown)


brockallen:
Closing as per [this](https://github.com/DuendeSoftware/Support/issues/113#issuecomment-1181768380).

PR #975, opened 2022-07-18
ignore duplicate param values in request processing

brockallen:
This PR adds logic when processing request parameters that are exact duplicates to filter/ignore them.

Fixes: https://github.com/DuendeSoftware/Support/issues/124

### DIFF
--- a/src/IdentityServer/Extensions/IReadableStringCollectionExtensions.cs
+++ b/src/IdentityServer/Extensions/IReadableStringCollectionExtensions.cs
@@ -7,6 +7,7 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Diagnostics;
+using System.Linq;
 
 #pragma warning disable 1591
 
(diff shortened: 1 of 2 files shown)


brockallen:
We used to do it this was back in IdentityServer4. But I noticed [here](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1) and [here](https://datatracker.ietf.org/doc/html/rfc6749#section-4.2.2.1) that dup parameters should treated as an `invalid_request` error (of course, I guess except for the resource parameter).

leastprivilege:
The RFC says

> Parameters sent without a value MUST be treated as if they were
   omitted from the request.  The authorization server MUST ignore
   unrecognized request parameters.  Request and response parameters
   MUST NOT be included more than once.

https://datatracker.ietf.org/doc/html/rfc6749#section-3.1

brockallen:
Will rework for 6.2 to validate against dups.

PR #1067, opened 2022-10-21
Fix PersistedGrantFilter.Validate()

svalbuena:
The PersistedGrantFilter.Validate() method verifies that the filter is not null and contains at least one filter value, but the way it's coded it does not allow the following:
* Only set the ClientId
* Only set the ClientIds
* Only set the Type
* Only set the Types

I think IdentityServer does not call the grant store with filters that follow the previous cases, that's why the Validate() call is not throwing exceptions right now. But the grant store implementations do contain the necessary logic (for example, see [this line](https://github.com/DuendeSoftware/IdentityServer/blob/50e71c72eb3905acc9fda4e0fa86c2510db665ea/src/IdentityServer/Stores/InMemory/InMemoryPersistedGrantStore.cs#L90)) to work on these cases, that's why I think that what I'm reporting is a bug.

I've added some integration tests to verify that the stores work fine in these four cases.

### DIFF
--- a/src/Storage/Extensions/PersistedGrantFilterExtensions.cs
+++ b/src/Storage/Extensions/PersistedGrantFilterExtensions.cs
@@ -18,12 +18,14 @@ public static class PersistedGrantFilterExtensions
     /// <param name="filter"></param>
     public static void Validate(this PersistedGrantFilter filter)
     {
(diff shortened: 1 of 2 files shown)


brockallen:
Thanks for this. Looks like we picked up this change already. Thanks!

PR #1074, opened 2022-10-25
Verify refresh token creation and lifetime in test

altenstedt:
Update the unit tests for refresh token to ensure that CreationTime and Lifetime properties do not change when renewed (TokenUsage.OneTimeOnly).

### DIFF
--- a/test/IdentityServer.UnitTests/Services/Default/DefaultRefreshTokenServiceTests.cs
+++ b/test/IdentityServer.UnitTests/Services/Default/DefaultRefreshTokenServiceTests.cs
@@ -287,6 +287,9 @@ public class DefaultRefreshTokenServiceTests
 
         oldToken.ConsumedTime.Should().Be(now);
         newToken.ConsumedTime.Should().BeNull();
+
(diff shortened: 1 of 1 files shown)


brockallen:
Hello, thanks for the PR. Since we don't have an open issue requesting this change, can you give us a little more context/info as to what prompted this addition? Was there something missing you uncovered?

Thanks!

altenstedt:
I need to adjust the expiration time for some of the refresh tokens in our database, based on the time they where created.  To do that, I wrote some code in our implementation of `IPersistedGrantStore`.

In my code, I took a dependency on the fact that `PersistedGrant.CreationTime` is set to the creation time for the original refresh token, even if you use `OneTimeOnly`.

It was not immediately clear to me from the documentation that this was guaranteed, so I thought that I would extend the aptly named unit test `...with_correct_dates` to assert this.  However, this is just a suggestion, so feel free to close this PR if it is not appropriate.  :-)

josephdecock:
This looks good to me. I've added a bit of detail to the docs about refresh token rotation as well.

josephdecock (approved):

Issue #545, opened 2021-11-29
Breaking change: Exception on adding custom claim type that already exists on token


ace90210:
**Which version of Duende IdentityServer are you using?**
 Duende IdentityServer 5.2.2
 
 Previously worked in Identity Server 3.1.3 and 4.1.2

**Which version of .NET are you using?**
Core 3.1

**Describe the bug**
When adding a custom claim that conflicts with existing claims on the token an exception occurs on line 109 of the TokenExtension.cs. This appears to be due to the fact previously the general "AddClaims" method was used which explicitly handles duplicate claims (by making them an array when detected). This was changed to manually checking and adding claims. The new approach does not check if the claim type is already in the dictionary before trying to add it.

**To Reproduce**

1. Create a client with client credentials flow and the custom claim below

2. `Claims = new List<ClientClaim>() { new ClientClaim("aud", "mycustomaudiencecheck") }`
 
3. request a token for this client and receive a 500 response

**Expected behaviour**
The same as previous behaviour. The existing claim is either converted to an array (if it was not an array already) and then the claim is added or the claim is added to the existing array of that type.

**Log output/exception with stacktrace**

```
System.ArgumentException: An item with the same key has already been added. Key: aud
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at Duende.IdentityServer.Extensions.TokenExtensions.CreateJwtPayloadDictionary(Token token, IdentityServerOptions options, ISystemClock clock, ILogger logger) in /_/src/IdentityServer/Extensions/TokenExtensions.cs:line 109
   at Duende.IdentityServer.Services.DefaultTokenCreationService.CreatePayloadAsync(Token token) in /_/src/IdentityServer/Services/Default/DefaultTokenCreationService.cs:line 85
   at Duende.IdentityServer.Services.DefaultTokenCreationService.CreateTokenAsync(Token token) in /_/src/IdentityServer/Services/Default/DefaultTokenCreationService.cs:line 76
   at Duende.IdentityServer.Services.DefaultTokenService.CreateSecurityTokenAsync(Token token) in /_/src/IdentityServer/Services/Default/DefaultTokenService.cs:line 292
   at Duende.IdentityServer.ResponseHandling.TokenResponseGenerator.CreateAccessTokenAsync(ValidatedTokenRequest request) in /_/src/IdentityServer/ResponseHandling/Default/TokenResponseGenerator.cs:line 426
[...]

PR #1075, opened 2022-10-25
add an integration test showing a custom claims service adding an audience claim

brockallen:


test/IdentityServer.IntegrationTests/Extensibility/CustomClaimsServiceTests.cs
```diff
@@ -0,0 +1,92 @@
...
+{
+    public CustomClaimsService(IProfileService profile, ILogger<DefaultClaimsService> logger) : base(profile, logger)
+    {
+    }
+
+    public override async Task<IEnumerable<Claim>> GetAccessTokenClaimsAsync(ClaimsPrincipal subject, ResourceValidationResult resourceResult, ValidatedRequest request)
+    {
+        var result = (await base.GetAccessTokenClaimsAsync(subject, resourceResult, request)).ToList();
+
+        result.Add(new Claim("foo", "foo1"));
```
ahaeber, line 88: Shouldn't the test be to add a `aud` claim, instead of a random `foo` claim? (Ref https://www.rfc-editor.org/rfc/rfc7519#section-4.1.3 for aud claim)

Seems to be related to https://github.com/DuendeSoftware/IdentityServer/issues/545 which was fixed by ignoring duplicate claims for `aud` and others.

brockallen: Well, the title of the PR is wrong. It was supposed to be a custom claim.

ahaeber: Ok, then it looks good to me

leastprivilege (approved):

Issue #1040, opened 2022-10-07
Check how EF7 handles task cancellations


Stra1ghter:
**Which version of Duende IdentityServer are you using?**
6.1.5
**Which version of .NET are you using?**
net6.0

**Describe the bug**  

When a user cancels a browser request that is handled by `IdentityServerMiddleware` the `TaskCanceledException` is logged as a critical log. We see the errors in different stack traces, but it is always a task that finally queries some store via EF Core.

**To Reproduce**

I couldn't absolutely reproduce the behaviour, but it [seems to be known](https://github.com/DuendeSoftware/IdentityServer/issues/541).
This usually always happens when you accept a `CancellationToken` in your actions, e.g. for long running operations, but don't handle the cancellation of the task.

Please note that it's now possible in EF Core to [distinguish between cancellation and failure](https://github.com/dotnet/efcore/issues/19526).

**Expected behavior**

Handle user caused task cancellations (e.g. clicking stop or refresh in the browser) by a custom filter, that could log this as information or debug.

Another solution would be to simply stop accepting and passing `CancellationToken` on IdentityServer's actions. I don't think IdentityServer has long running tasks that would benefit from knowing about the cancellation of a task, but I might be wrong here.

**Log output/exception with stacktrace**

```
Unhandled exception: "A task was canceled."
A task was canceled.
TaskCanceledExceptionSystem.Private.CoreLib   at Microsoft.EntityFrameworkCore.Storage.RelationalConnection.OpenInternalAsync(Boolean errorsExpected, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Storage.RelationalConnection.OpenInternalAsync(Boolean errorsExpected, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Storage.RelationalConnection.OpenAsync(CancellationToken cancellationToken, Boolean errorsExpected)
   at Microsoft.EntityFrameworkCore.Storage.RelationalCommand.ExecuteReaderAsync(RelationalCommandParameterObject parameterObject, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Query.Internal.SingleQueryingEnumerable`1.AsyncEnumerator.InitializeReaderAsync(AsyncEnumerator enumerator, CancellationToken cancellationToken)
[...]

PR #1084, opened 2022-11-03
Introduce filter to allow suppression of unhandled exception logs in IdentityServer middleware

brockallen:
Allows customers to decide if our catch block around any IdentityServer endpoints will emit a log entry when there is an unhandled exception. The default, and primary use case is to suppress logs when the HTTP connection is cancelled early (e.g. a user navigates away or cancelled the browser request) and EF internally is detecting this and throwing the TaskCancelledException.

Fixes: https://github.com/DuendeSoftware/IdentityServer/issues/1040


leastprivilege (approved):

brockallen:
@josephdecock needs docs :)

PR #1090, opened 2022-11-08
update from .NET 7 RC2 to .NET 7 RTM

brockallen:


leastprivilege (changes requested):
there's more to be done. See my PR

PR #1097, opened 2022-11-22
add exception to UnhandledExceptionEvent

raeffs:

When implementing a custom `EventSink` currently there is no access to the exception that was thrown when the event is of the type `UnhandledExceptionEvent`. The only way to get any information about the exception is to look at the string representation of the exception stored in the `Details` property of `UnhandledExceptionEvent`.

Access to the `Exception` type might be important, especially if the sink forwards the events to a framework / tool that supports exceptions. This is for example the case in certain logging frameworks. At the moment the default sink logs all the events with `_logger.LogInformation("{@event}", evt);`. In case of the `UnhandledExceptionEvent` it would be preferable to log it by providing the exception as separate parameter with `_logger.LogInformation(evt.Exception, "{@event}", evt);` so that the logging framework has access to it. At the moment it isn't possible to achieve this with a custom sink. With the change this would become possible.

### DIFF
--- a/src/IdentityServer/Events/UnhandledExceptionEvent.cs
+++ b/src/IdentityServer/Events/UnhandledExceptionEvent.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Duende Software. All rights reserved.
+// Copyright (c) Duende Software. All rights reserved.
 // See LICENSE in the project root for license information.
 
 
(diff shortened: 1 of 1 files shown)


brockallen:
@josephdecock To discuss.

brockallen (approved):

brockallen:
Looks good, thanks.

raeffs:
@brockallen I was wondering why this wasn't included in the 6.2.0 release? The release notes for 6.2.0 mention "A new option that can help filter unhandled exceptions out of the logs", I'm not sure if that refers to this change or something else.

brockallen:
hmm, maybe we just mislabeled this PR for the wrong milestone?

brockallen:
No, it looks like this PR was merged after we released 6.2.0.

raeffs:
ah, maybe I confused that with #1084 🙂

PR #1105, opened 2022-12-06
Refresh Token Consumption Rework

josephdecock:
This PR adds two new configuration options related to the handling of refresh tokens. The intent of both settings is to alleviate some of the database pressure associated with marking refresh tokens as consumed.

First, `DeleteOneTimeOnlyRefreshTokensOnUse` controls what happens to OneTimeUse tokens when they are used. They can now be either marked as consumed or deleted immediately. The idea here is that, if you aren't making use of the consumed tokens, you can safely delete them immediately. _The default is to immediately delete._

Second, `ConsumedTokenCleanupDelay` delays deletion of consumed tokens in the token cleanup job. The intent of the delay is to allow users to keep tokens for some period of time (perhaps a short delay to allow "lenient" one-time use refresh tokens that can be reused for a short interval, or perhaps a longer delay to allow for auditing, etc). _This interval defaults to 0._

We also should consider if we will change the defaults of `EnableTokenCleanup` and/or `RemoveConsumedTokens` - but perhaps changes to defaults belong in 7.0, rather than 6.3.

### DIFF
--- a/src/EntityFramework.Storage/Options/OperationalStoreOptions.cs
+++ b/src/EntityFramework.Storage/Options/OperationalStoreOptions.cs
@@ -87,6 +87,14 @@ public class OperationalStoreOptions
     public bool RemoveConsumedTokens { get; set; } = false;
 
     /// <summary>
+    /// Gets or sets the consumed token cleanup delay (in seconds). The default
+    /// is 0. This delay is the amount of time that must elapse before tokens
+    /// marked as consumed can be deleted. Note that only refresh tokens with
+    /// OneTimeOnly usage can be marked as consumed.
+    /// </summary>
+    public int ConsumedTokenCleanupDelay { get; set; } = 0;
+
+    /// <summary>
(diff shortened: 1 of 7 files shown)


josephdecock:
I do feel a little bit of tension between waiting for 7.0 to change defaults but also changing the default behavior (tokens are just deleted by default, not marked as consumed) in 6.3. But I think this is sort of a compromise - we're going to stop creating a bunch of consumed records by default in 6.3, and then in 7.0 by default we actually delete your old data.

brockallen:
These changes look good, but we also need to think about docs and release notes for this. Not sure where to collect those in the meantime.... @josephdecock

brockallen (approved):

josephdecock:
How about if I put draft release notes in the milestone description?

brockallen:
> How about if I put draft release notes in the milestone description?

Maybe just open a new issue and do it there?

josephdecock:
Issue for 6.3 release notes is here: https://github.com/DuendeSoftware/IdentityServer/issues/1136

PR #1118, opened 2023-01-06
Fixes ToOptimizedRawValues to handle multiple resource values when combined with resource indicators

runegri:
When combining Resource Indicators with a Request Object the extension method ValidatedAuthorizeRequest.ToOptimizedRawValues creates a new NameValueCollection that holds any values that are not represented in the RequestObject. This method uses the indexer of the original NameValueCollection, but that has the unfortunate effect that repeated keys are represented as a comma separated string. 

This means that when supplying multiple resource indicators they are merged into one incorrect value. For instance the resource indicators ["Resource1", "Resource2"] are merged into the value "Resource1,Resource2". This will obviously fail at a later time when validating the resource indicators, specifically this fails after doing an external logon and we call the OidcReturnUrlParser. 

This pull request changes the setup of the new NameValueCollection so that repeated keys are handled as individual values by iterating through all values for the given key. 

A unit test that verifies the fix is included.


brockallen:
There's a reason we don't do it that way --- I'll have to go back in the history to remember why.

brockallen (approved):

josephdecock (changes requested):
Let's hold off on merging to main until we decide if we are patching 6.2.

runegri:
Great! 
None of our users have reported the issue yet, but we are heavy users of request objects and more or less mandate the use of resource indicators for most apis. So it is only a matter of time before someone hits it. 
So a patch would be preferable, but you guys decide on what makes most sense

josephdecock (approved):

brockallen:
We released 6.2.2 yesterday with this update. Let us know how it works for you. Thanks.

runegri:
Thanks for this one! We are pushing it into our public test environment now :)

PR #1134, opened 2023-01-19
add support for unmet_authentication_requirements

### LINKED ISSUE #1133: Add support for unmet_authentication_requirements
For: https://datatracker.ietf.org/doc/html/draft-ietf-oauth-step-up-authn-challenge

To AuthorizationError enum, AuthorizeResult processing, and AuthorizeInteractionResponseGenerator.

brockallen:
This adds support for the new unmet_authentication_requirements added in "OAuth 2.0 Step-up Authentication Challenge Protocol": https://datatracker.ietf.org/doc/html/draft-ietf-oauth-step-up-authn-challenge

Closes #1133


brockallen:
For docs I don't know if we currently doc the error modes, but if we do we should add the new value.

josephdecock (approved):

josephdecock:
> For docs I don't know if we currently doc the error modes, but if we do we should add the new value.

We don't currently describe the error modes. New issue added to address that.

Issue #432, opened 2021-09-24
Consider supporting the OIDC prompt=create param


brockallen:
https://openid.net/specs/openid-connect-prompt-create-1_0.html

I think we currently restrict the allowed set of prompt params to be the ones from the core spec. This would be a small modification to expand the allowed list internally. I don't think we'd do anything on the UI side, unless we wanted to introduce a new UI endpoint for registration (much like we might for "select account")?

g7ed6e:
would you accept a PR with the below modifications ?
- add `prompt_values_supported` to the discovery endpoint
- add UI endpoint
- add redirection to the new UI endpoint when `prompt=create` is supplied to the authorize endpoint

brockallen:
Not at this time, but thank you.

g7ed6e:
Now that IS6 is released would you reconsider ? I’m migrating from IS4v4 to IS5/6 and would be happy to drop a custom IAuthorizeInteractionResponseGenerator.

brockallen:
Yes, we will consider this request in our next round of planning. Thx.

g7ed6e:
Great i will submit a PR. Thanks

brockallen:
> Great i will submit a PR. Thanks

What I meant is that we'll consider it internally in our product planning. Please don't spend any time until we have confirmed that we want to add it - I don't want you to waste time if we are unsure we are adding (and supporting) the feature. Thanks.

g7ed6e:
@brockallen you can assign to me

leastprivilege:
We added this to our next milestone - which means we will definitely look into it. 

For us it is typically easier and less time consuming to just do the work ourselves than reviewing, testing, discussing and integrating external PRs. Thanks for the offer though!!!

brockallen:
The general redirect response processing could use a revamp, but would cause a breaking change. If we support this, then I'd want to do it as part of that revamp.

brockallen:
@g7ed6e, @filipw

When you use prompt=create today, do you ever use it with other prompt values? The spec only recommends you not do this, but I'm curious how you use it.

Also, @g7ed6e suggested adding a new property on the InteractionResponse to indicate create, and this implies a new URL on the UserInteraction. I wasn't going to do that in this version, and save major rework on the response processing until a major release. So this would mean for now we'd just redirect to the login URL with the prompt=create.

Thoughts?

g7ed6e:
> @g7ed6e, @filipw
> 
> 
> 
[...]

PR #1153, opened 2023-01-31
Add support for OIDC prompt=create

brockallen:
This PR adds support for the OIDC prompt=create extension:

https://openid.net/specs/openid-connect-prompt-create-1_0.html

Closes: https://github.com/DuendeSoftware/IdentityServer/issues/432

Docs needed: New CreateAccountUrl property on the UserInteractionOptions, and how this affects the values in discovery and request processing.

### DIFF
--- a/src/IdentityServer/ResponseHandling/Models/InteractionResponse.cs
+++ b/src/IdentityServer/ResponseHandling/Models/InteractionResponse.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Duende Software. All rights reserved.
+// Copyright (c) Duende Software. All rights reserved.
 // See LICENSE in the project root for license information.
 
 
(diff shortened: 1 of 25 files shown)


src/IdentityServer/ResponseHandling/Models/InteractionResponse.cs
```diff
@@ -1,16 +1,48 @@
...
 
 
 using Duende.IdentityServer.Extensions;
 
 namespace Duende.IdentityServer.ResponseHandling;
 
+/// <summary>
+/// Models the types of interaction results from the IAuthorizeInteractionResponseGenerator
+/// </summary>
+public enum InteractionResponseType
```
josephdecock, line 12: Would it make sense to put Login, Consent, and Create into this enum?

brockallen: The point was to have more of a summary, rather than specific flags to check.

brockallen: Maybe there's a better way to model this, tho?

josephdecock: We looked at this together, but didn't like the alternatives - we want to keep existing stuff for backwards compat, and can't cleverly relate the old and new (the collection of booleans might be a weird mixture, and what then?)

josephdecock (approved):

Issue #584, opened 2021-12-11
Consider adding a ready to use health check endpoint


leastprivilege:
e.g. checking disco endpoint for in-mem and doing a db connectivity check for EF...

brockallen:
We don't need to add anything really to support this style of health check. Just loading the endpoints and using them via our public API is sufficient.

PR #1154, opened 2023-01-31
fix non-unique names in endpoint list and add health check sample

brockallen:
This PR contains two things:
* bug fix: the JWKS Endpoint in the endpoint collection was using the same name as the discovery endpoint, so the lookup by name results in an ambiguous/non-unique result
* sample to show how to use our public APIs to create a health check for the discovery endpoint

Closes: https://github.com/DuendeSoftware/IdentityServer/issues/584


brockallen:
@leastprivilege do we then also want docs or sample, or is this sufficient?

leastprivilege (approved):

PR #1157, opened 2023-02-02
Always emit the sid claim in id tokens

josephdecock:
In the past, we only needed the sid claim in the check session endpoint, and only included the sid claim when that endpoint was enabled check session. Now we have more cases where the sid claim is useful, so we are no longer going to make the sid claim conditional on the EnableCheckSessionEndpoint option

### DIFF
--- a/src/IdentityServer/Validation/Default/AuthorizeRequestValidator.cs
+++ b/src/IdentityServer/Validation/Default/AuthorizeRequestValidator.cs
@@ -874,26 +874,23 @@ internal class AuthorizeRequestValidator : IAuthorizeRequestValidator
         }
 
         //////////////////////////////////////////////////////////
-        // check session cookie
+        // session id
         //////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


src/IdentityServer/Validation/Default/AuthorizeRequestValidator.cs
```diff
@@ -874,27 +874,24 @@ private async Task<AuthorizeRequestValidationResult> ValidateOptionalParametersA
...
-                else
-                {
-                    LogError("Check session endpoint enabled, but SessionId is missing", request);
-                }
+                request.SessionId = sessionId;
             }
             else
             {
-                request.SessionId = ""; // empty string for anonymous users
+                LogError("SessionId is missing", request);
```
josephdecock, line 888: I don't think it is normally possible to not have a session id, so I think logging this as an error and then continuing is reasonable.

brockallen (approved):

Issue #1169, opened 2023-02-06
Backchannel Logout Token Issues


Cephei:
**Which version of Duende IdentityServer are you using?**
6.2.2

**Which version of .NET are you using?**
6

**Describe the bug**

1. Logout tokens do not use the recommended header type `logout+jwt`. It looks like IdentityServerTools may need to be refactored a bit to provide the token type for the `Token` class constructor. See: https://openid.net/specs/openid-connect-backchannel-1_0.html#LogoutToken
2. ~~It appears like our Logout Tokens are being signed properly with the private key, however, the Logout Token JWT algorithm shows it as a symmetric HS256. I believe this is causing validation to fail, however, perhaps I'm misunderstanding something. See the screenshot and below example token. The json/certificate in the screenshot comes directly from our Duende `jwks` endpoint.~~ **UPDATE:** Not a bug.

**Example JWT**
```
eyJhbGciOiJIUzI1NiIsImtpZCI6IjA3NkYxOThBNjhDQjVDN0ZGREFFOTY2QkY5MDcwMTFEIiwieDV0IjoiZE9aQW5qUHFRVnVXclZLbmluLS1HcTQzSkFZIiwidHlwIjoiYXQrand0In0.eyJpc3MiOiJodHRwczovL2FjY291bnRzLnFhLnNvbWVleGFtcGxlb3JnLm5ldCIsIm5iZiI6MTY3NTM2MjkxOSwiaWF0IjoxNjc1MzYyOTE5LCJleHAiOjE2NzUzNjMyMTksImF1ZCI6InRpbWUtbG9jYWwiLCJqdGkiOiIwNzI3QkFBNkJGMDJDREU1RkEyRTRDQTQ3Njc3MEIzNiIsImV2ZW50cyI6eyJodHRwOi8vc2NoZW1hcy5vcGVuaWQubmV0L2V2ZW50L2JhY2tjaGFubmVsLWxvZ291dCI6e319LCJzdWIiOiI4NCIsInNpZCI6IjVBMkUxNURDRDI3RENFNTY4MTEwQjJEOTFFRjA0MTZCIn0.X0XZYFFZsehhjO0IuEceFEXsWN7uqF2TLmHhuBLhUeo
```
<img width="1245" alt="Screen Shot 2023-02-05 at 9 45 38 PM" src="https://user-images.githubusercontent.com/1054639/216873144-41d91bd4-0b8c-451c-8271-6409dbf03c2b.png">

[...]

PR #1170, opened 2023-02-06
add LogoutTokenJwtType property for back channel logout tokens

brockallen:
This adds a new `LogoutTokenJwtType` option for explicitly configuring back channel logout tokens `typ` header. It defaults to `logout+jwt` to be in compliance with the recently updated and released OIDC back channel logout spec.

https://openid.net/specs/openid-connect-backchannel-1_0.html#LogoutToken

While this might be a breaking change, the current versions of the .NET JWT handler ignore the `typ` so it is a low chance of affecting customers. And if they are validating the `typ` then they should be using the value as per spec. If this does cause issues, then they can set the new `LogoutTokenJwtType` option value introduced in this PR.

Closes: https://github.com/DuendeSoftware/IdentityServer/issues/1169

### DIFF
--- a/src/IdentityServer/Configuration/DependencyInjection/Options/IdentityServerOptions.cs
+++ b/src/IdentityServer/Configuration/DependencyInjection/Options/IdentityServerOptions.cs
@@ -34,6 +34,14 @@ public class IdentityServerOptions
     public string AccessTokenJwtType { get; set; } = "at+jwt";
 
     /// <summary>
+    /// Gets or sets the value for the JWT typ header for logout tokens.
+    /// </summary>
+    /// <value>
+    /// The JWT typ value.
+    /// </value>
+    public string LogoutTokenJwtType { get; set; } = "logout+jwt";
+
+    /// <summary>
     /// Emits an aud claim with the format issuer/resources. That's needed for some older access token validation plumbing. Defaults to false.
     /// </summary>
     public bool EmitStaticAudienceClaim { get; set; } = false;
--- a/src/IdentityServer/IdentityServerConstants.cs
+++ b/src/IdentityServer/IdentityServerConstants.cs
@@ -51,6 +51,7 @@ public static class IdentityServerConstants
     {
         public const string IdentityToken = "id_token";
         public const string AccessToken = "access_token";
+        public const string LogoutToken = "logout_token";
     }
 
     public static class ClaimValueTypes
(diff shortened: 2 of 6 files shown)


brockallen:
Hey @josephdecock -- quick fix for the back channel logout issue. We can discuss/review together for what release to add this into.

brockallen:
TODO:

* look thru code and fix where we use IdentityModel.OidcConstants and ensure those are only used as protocol params.
* update IdentityModel if needed to add the back channel protocol constants

brockallen:
> While this might be a breaking change, the current versions of the .NET JWT handler ignore the typ so it is a low chance of affecting customers. And if they are validating the typ then they should be using the value as per spec. If this does cause issues, then they can set the new LogoutTokenJwtType option value introduced in this PR.

Add this as a big point in the release/upgrade notes.

brockallen:
> look thru code and fix where we use IdentityModel.OidcConstants and ensure those are only used as protocol params.

@leastprivilege turns out this is a much larger chore, as OidcConstants.TokenTypes is used in a lot of places.

brockallen:
> update IdentityModel if needed to add the back channel protocol constants

We do not -- they all seem covered already.

josephdecock:
> > look thru code and fix where we use IdentityModel.OidcConstants and ensure those are only used as protocol params.
> 
> @leastprivilege turns out this is a much larger chore, as OidcConstants.TokenTypes is used in a lot of places.

I think we're fine here. My search for OidcConstants.TokenTypes shows that we only ever use these constants as a parameter to the the constructor for Token or to compare to the Token's type parameter. There are some instances where the Token class isn't directly being used for a protocol message (e.g., events) but it seems fine to me.

josephdecock (approved):

LeishaBurford:
Turns out this did break my legacy net4.8 client because it _was_ checking the `typ` of the token and now refuses to read it

```
 System.IdentityModel.Tokens.SecurityTokenException: IDX10702: Jwt header type specified, must be 'JWT' or 'http://openid.net/specs/jwt/1.0'.  Type received: 'logout+jwt'.
   at System.IdentityModel.Tokens.JwtSecurityToken.Decode(String jwtEncodedString)
   --- End of inner exception stack trace ---
   at System.IdentityModel.Tokens.JwtSecurityToken.Decode(String jwtEncodedString)
   at System.IdentityModel.Tokens.JwtSecurityTokenHandler.ReadToken(String tokenString)
   at System.IdentityModel.Tokens.JwtSecurityTokenHandler.ValidateSignature(String token, TokenValidationParameters validationParameters)
   at System.IdentityModel.Tokens.JwtSecurityTokenHandler.ValidateToken(String securityToken, TokenValidationParameters validationParameters, SecurityToken& validatedToken)
```

Easy to fix in my identity server, PR description helped. Just wanted to note that it was in fact breaking for at least one person 😁

Issue #1171, opened 2022-12-14
System.ArgumentOutOfRangeException in DynamicSchemeAuthenticationMiddleware


vitalyobukhov:
**Which version of Duende IdentityServer are you using?**
6.1.0

**Which version of .NET are you using?**
net6.0

**Describe the bug**
`Duende.IdentityServer.Hosting.DynamicProviders.DynamicSchemeAuthenticationMiddleware`  throws  `System.ArgumentOutOfRangeException` in case if `/scheme` part was not provided within the URI of request.

**To Reproduce**
Just plug in the standard IdentityServer via `IdentityServerApplicationBuilderExtensions.UseIdentityServer` within the web app and try to access `/federation` path without trailing slash or scheme-specific suffix.

**Expected behavior**
No exception - avoid `context.Request.Path.Value.Substring(startIndex + 1)` if `startIndex >=  context.Request.Path.Value.Length`.

**Log output/exception with stacktrace**
```
System.ArgumentOutOfRangeException
startIndex cannot be larger than length of string. (Parameter 'startIndex')
   at System.String.Substring(Int32 startIndex, Int32 length)
   at Duende.IdentityServer.Hosting.DynamicProviders.DynamicSchemeAuthenticationMiddleware.Invoke(HttpContext context) in /_/src/IdentityServer/Hosting/DynamicProviders/DynamicSchemes/DynamicSchemeAuthenticationMiddleware.cs:line 48
```

**Additional context**
Found it during ISEs 500 triage.

brockallen:
PR submitted for this, @vitalyobukhov.

PR #1172, opened 2023-02-06
requests to ~/federation with no additional segments should return not found

brockallen:
Fixes: https://github.com/DuendeSoftware/IdentityServer/issues/1171


brockallen:
@josephdecock maybe there's a better check, but I simply went with a length check.

josephdecock (approved):
> @josephdecock maybe there's a better check, but I simply went with a length check. 

Nah, seems nice and simple

PR #1176, opened 2023-02-08
Use correct sensitive values filter during CIBA

josephdecock:
Use the BackchannelAuthenticationRequestSensitiveValuesFilter in the BackchannelAuthenticationRequestValidator.

### DIFF
--- a/src/IdentityServer/Configuration/DependencyInjection/Options/LoggingOptions.cs
+++ b/src/IdentityServer/Configuration/DependencyInjection/Options/LoggingOptions.cs
@@ -22,7 +22,6 @@ public class LoggingOptions
     public ICollection<string> BackchannelAuthenticationRequestSensitiveValuesFilter { get; set; } =
         new HashSet<string>
         {
-            // TODO: IdentityModel
             OidcConstants.TokenRequest.ClientSecret,
             OidcConstants.TokenRequest.ClientAssertion,
             OidcConstants.AuthorizeRequest.IdTokenHint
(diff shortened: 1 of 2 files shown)


src/IdentityServer/Configuration/DependencyInjection/Options/LoggingOptions.cs
```diff
@@ -22,7 +22,6 @@ public class LoggingOptions
     public ICollection<string> BackchannelAuthenticationRequestSensitiveValuesFilter { get; set; } =
         new HashSet<string>
         {
-            // TODO: IdentityModel
```
josephdecock, line 25: Review: Unclear what this TODO was for.

src/IdentityServer/Configuration/DependencyInjection/Options/LoggingOptions.cs
```diff
@@ -22,7 +22,6 @@ public class LoggingOptions
...
             OidcConstants.TokenRequest.ClientSecret,
             OidcConstants.TokenRequest.ClientAssertion,
             OidcConstants.AuthorizeRequest.IdTokenHint
```
josephdecock, line 27: I think this is enough of a filter, but I imagine some organizations might also want to remove login_hint and login_hint_token, because maybe email address is used as the username, and email addresses are PII, thus not allowed in logs. I think letting people add that to the filter config if they need it is fine. Thoughts?

src/IdentityServer/Configuration/DependencyInjection/Options/LoggingOptions.cs
brockallen: Likely we were waiting on a constant to be defined and released in a new version of IdentityModel.

brockallen (approved):

PR #1178, opened 2023-02-08
Add TokenResponseGenerator.CreateResponseAsync

josephdecock:
CreateResponseAsync is a new virtual method in the
TokenResponseGenerator. CreateResponseAsync does the same thing that
ProcessTokenRequestAsync used to do (generate access and refresh tokens
and add them to a response object), but it is called by all token flows
except the refresh token flow, which generates its tokens differently.

TokenResponseGenerator.ProcessTokenRequestAsync is now an alias for
CreateResponseAsync and is deprecated in favor of it, with the intention
to completely remove ProcessTokenRequestAsync in 7.0.


brockallen:
Ok, so let's do this:
1) yes, update all the duplicated code to call into the single helper API
2) no, let's leave the name of the old API the same (for the backwards compat reasons previously discussed)

brockallen (approved):

brockallen:
@josephdecock the above text doesn't quite match anymore.

Issue #1116, opened 2023-01-03
Add DPoP support


leastprivilege:
final draft:
https://www.ietf.org/archive/id/draft-ietf-oauth-dpop-12.html

Issue #1059, opened 2022-10-18
review ICache<T> and IDistribiutedCache


brockallen:
We have two different uses of caching. Perhaps there's some cleanup/consolidation that we could do.

brockallen:
Quick review here -- ICache\<T> is generally used for caching data returned from stores, and IDistributedCache is used for more temp/operational/volatile data such as replay caches, request params, and throttling services.

brockallen:
@josephdecock any input here as to suggested direction? IOW, do we see that people want some improvement here? or maybe it's just more to document what's used where? also, i could see an adding an ICache\<T> based on IDistributedCache. 

Anyway, we should discuss.

brockallen:
For reference if useful (not sure that it is): https://devblogs.microsoft.com/dotnet/caching-abstraction-improvements-in-aspnetcore/

josephdecock:
I don't hear much confusion about the two interfaces. Whenever production deployments/scaling comes up, I always mention that IDistributedCache might be necessary, and our docs on how we use it have seen some recent improvements. I think we could add a bit of clarification to them to point out how it is distinct from the ICache. 

I also don't think there would be a big enough benefit to justify making a breaking changing here.

gpolcode:
@brockallen we would like to see an improvement here.
We are using the provided Caching Stores in a multi regional architecture. 
The current caching setup has a few disadvantages for our usecase.

The default implementation of ICache<T> uses IMemoryCache. This could lead differences between regions. A provided implementation of ICache<T> using IDistributedCache could solve this issue.

A client may change, however the Caching Stores have no API to invalidate a cache entry. We could only guess/copy what key the Stores are using.

Is there already a solution to these issues I may have missed?

ivannedyalkov:
it is not trivial to implement. However, IS is interface dependent. In our case, it's quite a complex k8s deployment with multiple pods in different regions, and Memory Cache is simply not a solution. If you implement ICache<T> you can use whatever you want under it. For DI, just use your distributed implementation.

`builder.Services.AddTransient(typeof(ICache<>), typeof(WhateverCache<>));`

However, it is not well documented and requires a lot of digging around in the code. @josephdecock @brockallen

brockallen:
Ok so then we can consider additional docs here. Thanks!

mrjamiebowman:
[...]

PR #1184, opened 2023-02-17
First draft of DPoP support

brockallen:
First cut of DPoP support (without a built-in nonce implementation).

Open issues:
* Need an update for the IdentityModel constants
* For clients that are configured to requite DPoP do we require the dpop_jkt param on the authorize endpoint? The spec says OPTIONAL, and that flag will still[ require the client to provide one on the token endpoint.
* We need to think thru the options for replay duration and where those live. Also, this will require the replay cache by default now in DI, which requires an IDistributedCache. (probably related: https://github.com/DuendeSoftware/IdentityServer/issues/1059)
* We need to think thru the options for proof token iat duration validation and where those live. Related: do we want iat vs nonce validation to be a per-client setting?
* Still need a design for nonce validation and/or extensibility. 
* Should we support dpop_jkt with CIBA on the back channel authentication request?
* Needs docs and samples.

Closes: https://github.com/DuendeSoftware/IdentityServer/issues/1116

### DIFF
--- a/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/Additional.cs
+++ b/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/Additional.cs
@@ -336,7 +336,6 @@ public static class IdentityServerBuilderExtensionsAdditional
     /// <returns></returns>
     public static IIdentityServerBuilder AddJwtBearerClientAuthentication(this IIdentityServerBuilder builder)
     {
-        builder.Services.TryAddTransient<IReplayCache, DefaultReplayCache>();
         builder.AddSecretParser<JwtBearerClientAssertionSecretParser>();
         builder.AddSecretValidator<PrivateKeyJwtSecretValidator>();
 
--- a/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/Core.cs
+++ b/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/Core.cs
@@ -215,7 +215,6 @@ public static class IdentityServerBuilderExtensionsCore
         builder.Services.TryAddTransient<IUserCodeGenerator, NumericUserCodeGenerator>();
         builder.Services.TryAddTransient<ILogoutNotificationService, LogoutNotificationService>();
         builder.Services.TryAddTransient<IBackChannelLogoutService, DefaultBackChannelLogoutService>();
-        builder.Services.TryAddTransient<IResourceValidator, DefaultResourceValidator>();
         builder.Services.TryAddTransient<IScopeParser, DefaultScopeParser>();
         builder.Services.TryAddTransient<ISessionCoordinationService, DefaultSessionCoordinationService>();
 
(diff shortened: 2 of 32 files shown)


brockallen:
@josephdecock and @leastprivilege we should schedule a time to review this first cut, and all together seems to make sense.

leastprivilege:
> For clients that are configured to requite DPoP do we require the dpop_jkt param on the authorize endpoint? The spec says OPTIONAL, and that flag will still[ require the client to provide one on the token endpoint.

In FAPI 2 the `dpop_jkt` is mandatory. So this needs a separate `Require...` setting.

brockallen:
> In FAPI 2 the dpop_jkt is mandatory. So this needs a separate Require... setting.

[This](https://openid.bitbucket.io/fapi/fapi-2_0-security-profile.html#section-5.3.1.1) seems to say: 

> if using DPoP, shall support "Authorization Code Binding to DPoP Key" (as required by section 10.1 of [[I-D.ietf-oauth-dpop](https://openid.bitbucket.io/fapi/fapi-2_0-security-profile.html#I-D.ietf-oauth-dpop)]).

We do support it, so I'm not sure the above is worded such that we need a new client config just for that aspect. Thoughts?

leastprivilege:
Ok. I misread that.

brockallen:
> For clients that are configured to requite DPoP do we require the dpop_jkt param on the authorize endpoint? The spec says OPTIONAL, and that flag will still[ require the client to provide one on the token endpoint.

Ok, so I think the answer is that we don't need a require flag for this.

brockallen:
> We need to think thru the options for replay duration and where those live. Also, this will require the replay cache by default now in DI, which requires an IDistributedCache. (probably related: https://github.com/DuendeSoftware/IdentityServer/issues/1059)

I added a new DPoPOptions on the IdentityServer options with the validity duration. I guess we also want a clock skew option as well? This might be necessary even for server-generated nonce values.

brockallen:
Also added a simple nonce implementation that data protects the server time and uses that for the expiration checking. Still need a flag somewhere to trigger this. It feels like it should be per-client.

brockallen:
> I guess we also want a clock skew option as well? This might be necessary even for server-generated nonce values.

This has now been added too.

brockallen:
> We need to think thru the options for proof token iat duration validation and where those live. Related: do we want iat vs nonce validation to be a per-client setting? I can see a simple built-in nonce implementation that simply returns a DP protected server timestamp. It'd really not be too diff than the iat validation, but it's server-generated. I can see 2 per-client settings: 1) DPoPProofTokenLifetime (or somesuch), and 2) DPoPValidationType (enum iat vs nonce). Our built-in can do both, I think.

I added a per-client setting for this. To be reviewed/discussed.

brockallen:
We still need an update/release of IdentityModel. I think all the PRs with the updated constants have been merged.

brockallen:
Added MVC sample that works with both the OIDC handler and our access token management library. Yet to add API support.

brockallen:
Docs topics:
* new client config settings
* use of IReplayCache
* update authZ and token EP docs with new params

brockallen:
TODO: look at token renewal proof key/cnf binding logic.

leastprivilege (approved):
LGTM

Issue #111, opened 2021-02-17
Add dynamic client registration


leastprivilege:
https://tools.ietf.org/html/rfc7591
https://openid.net/specs/openid-connect-registration-1_0.html

Cephei:
Would really love to see this built out to support 3rd-Party Initiated Login
https://openid.net/specs/openid-connect-core-1_0.html#ThirdPartyInitiatedLogin

leastprivilege:
What exactly are you looking for?

dynamic client registration
or
3rd init login

Because both are completely independent and one is possible without the other.

Cephei:
Looking for dynamic client registration. I think the only relation that seems to be missing in IdSrv is the built-in support for `initiate_login_uri` on `Client`. I've been prototyping migrating away from Auth0 to IdSrv Enterprise for my company and the org that recently acquired us.

To have a seamless integration for our mutual customers, 3rd-party-init seems like the best approach for us to achieve SSO to switch between platforms; as it functionally resembles SAML's IdP-Initiated flow, but without the CSRF vulnerability issues.

leastprivilege:
OK. I see. Well - as of today there is no dynamic client registration. We might add this for v6. That being said, you can add this yourself on top of your data store today. 

Is this blocking you right now?

anderson-santana-ff:
Hi, @leastprivilege ! Hope you are well.

Any news and documentation about DCR?

leastprivilege:
No - not right now.

What exactly are you looking for?

anderson-santana-ff:
Thank you for replying. Looking for dynamic client registration or a guide/documentation to extend it in current Identity Server (6.0) to have it. If you can share any guidance would great :)

leastprivilege:
We might be adding it as part of the config API
https://blog.duendesoftware.com/posts/20220204_admin_ui/

In the meantime, you could simply create an API endpoint that accepts a DCR document and talk to the config database yourself.

https://openid.net/specs/openid-connect-registration-1_0.html#RegistrationRequest

No1e:
Hi. Is this definately planned for 6.2.0?

Best regards,
Novak

leastprivilege:
no - this is our current thinking but might change.

leastprivilege:
Having a closer look - it would make sense to start with implementing an applicable subset of the DCR protocol to start with
https://datatracker.ietf.org/doc/html/rfc7591

[...]

PR #1206, opened 2023-03-13
First cut of DCR

josephdecock:
First cut of the new configuration projects and dynamic client registration support.


josephdecock:
Linking to #111, but keeping issue open as there is still more to do before release.

brockallen:
So far so good. The TODOs file is a good idea to track pending work. Thanks!

brockallen (approved):

PR #1209, opened 2023-03-15
Make ConfigurationDbContext overridable

f-waffenschmidt:
To inherit from ConfigurationContext it is easier to use the base class of DbContextOptions without generic like already done in the PersistedGrantDbContext. This makes it easier to add additional DbSets<>, overrides to the context. Without you'll probably need to register the generic overload by yourself to make it injectable to the constructor or use the interface and write an own database context inherited from IConfigurationDbContext

### DIFF
--- a/src/EntityFramework.Storage/DbContexts/ConfigurationDbContext.cs
+++ b/src/EntityFramework.Storage/DbContexts/ConfigurationDbContext.cs
@@ -50,7 +50,7 @@ public class ConfigurationDbContext<TContext> : DbContext, IConfigurationDbConte
     /// </summary>
     /// <param name="options">The options.</param>
     /// <exception cref="ArgumentNullException">storeOptions</exception>
-    public ConfigurationDbContext(DbContextOptions<TContext> options)
+    public ConfigurationDbContext(DbContextOptions options)
         : base(options)
     {
     }


brockallen:
Thanks for the PR, but I don't think this will work when you have multiple DbContextOptions in the DI system at the same time. This is why we have the constraint in the first place.

brockallen:
If there's some other approach you can think of to achieve what you're looking for, we're open to that discussion. So let us know if there's some other way you can think that it could work.

f-waffenschmidt:
You're right, it is not working with the same DbContextOptions with many different contexts. I just thought it would be easier to add more DbSets to the context itself or even some classes to extend with own properties. Like the same approach AspNetIdentity uses.

What I have in mind is something like this: 

```csharp
/// <summary>
/// Base ConfigurationContext with generic overload for client extension
/// </summary>
/// <typeparam name="TClient"></typeparam>
public class ConfigurationDbContext<TClient> : Duende.IdentityServer.EntityFramework.DbContexts.ConfigurationDbContext
    where TClient : Client
{
    public ConfigurationDbContext(
        DbContextOptions<Duende.IdentityServer.EntityFramework.DbContexts.ConfigurationDbContext> options) :
        base(options)
    {
    }

    protected override void OnModelCreating(ModelBuilder modelBuilder)
    {
        StoreOptions = new ConfigurationStoreOptions();
        base.OnModelCreating(modelBuilder);
    }

    public DbSet<TClient> Clients { get; set; }
}
```
 
Now it would be possible to add my own implementation to the client by overriding the dbContext

```csharp
/// <summary>
/// Custom implementation to extend base DbSets with custom client implementation
/// </summary>
public class CustomConfigurationDbContext : ConfigurationDbContext<ExtendedClient>
{
    public CustomConfigurationDbContext(
        DbContextOptions<Duende.IdentityServer.EntityFramework.DbContexts.ConfigurationDbContext> options) :
        base(options)
    {
    }

    /// <summary>
    /// A new dbset used to assign tenants to clients
    /// <summary>
    public DbSet<ClientTenant> ClientTenants { get; set; }

    protected override void OnModelCreating(ModelBuilder modelBuilder)
    {
        modelBuilder.Entity<ExtendedClient>()
            .HasOne(bc => bc.Tenant)
            .WithMany(b => b.Clients)
            .HasForeignKey(bc => bc.TenantId);
        base.OnModelCreating(modelBuilder);
    }
}

public class ClientTenant
{
    /// <summary>
    /// Identifier for the tenant
    /// </summary>
    public string TenantName { get; set; }

    public IEnumerable<ExtendedClient> Clients { get; set; }
}

/// <summary>
/// Extended Client to get the client related to a tenant on database level
/// </summary>
public class ExtendedClient : Client
{
    public Guid TenantId { get; set; }
    public ClientTenant Tenant { get; set; }
}
```

Instead having a database with all clients as a global list I could easily filter by the tenant on database level. Of course this approach would lead to other changes, like the way a client is identified during token flow. What comes to my mind is the JWT Bearer client assertion where I could add a tenant to the signed token request. 

But this might be independent of the generic-less DbContextOptions.

brockallen:
@josephdecock -- any ideas?

brockallen:
reopening to not forgot to review/brainstorm

josephdecock:
I'm closing this PR because I don't think there is an alternative that we don't already have in our our existing extensibility points.

If you want to have tenancy used deeply throughout the store, you probably need to write a custom store that can properly make use of tenancy information.

If you just want to attach some tenancy information to the existing entities, I don't see any reason why you couldn't use our entities (or extended versions of our entities) in a new db context.

PR #1212, opened 2023-03-15
Add possibility to iterate through a list of configured event sinks. 

f-waffenschmidt:

This PR allows to step through multiple configured event sinks in the DefaultEventService. For instance writing events to windows event log or to databases. I've added the windows event log sink. As an idea it would be nice to write the events into the grants db as another persistent storage besides this event log implementation. 

***Motiviation***
The use case is very simple. To support audits or detect at a attacks as early as possible administrators of the software might not have enough knowledge to get into the log output of the application. Furthermore administrators will see lot of events depending on the configured LogLevel. To separate common logging from events other sinks could be used. Without this implementation library users need to override the the the DefaultEventSink completely and duplicate code. 

***Prerequisite***
To write data into event log, the executed service must either run as administrator or an registry key must be added.

***Examples***
image
image


brockallen:
Hi @f-waffenschmidt. I think you can already achieve what you want by extending our base classes and registering your own in DI, thus overriding our defaults. Have you tried that yet?

f-waffenschmidt:
Hi  @brockallen,

yes it is possible and it is working as expected. It is nothing what I need or what blocks me. I just wanted to share a maybe more convenience way to write to multiple event sinks via DI registration. 

Right now something like this is necessary:

```csharp
public class MyCustomEventSink : DefaultEventSink
{
    public MyCustomEventSink(ILogger<DefaultEventService> logger) : base(logger)
    {
        
    }

    public override Task PersistAsync(Event evt)
    {
        if (auditIsActive)
        {
            //Write to eventlog
        }

        if (persistenceRequired)
        {
            //Write to database
        }
        //...
    
        return base.PersistAsync(evt);
    }
}
```

But as I said, it works

Thanks

brockallen:
Ok, thanks for the reply. We'll review this to see if any of it makes sense for us to include. Of course, anything we include means we then need to maintain it, so there's always that balance and trade off to consider. Thanks.

f-waffenschmidt:
Yes, of course. The event log implementation is quite custom for a dedicated OS and database might be the same depending on what you want to achieve as developer or what the requirement is.

When I analyzed your code I saw similar implementations, like [here](https://github.com/DuendeSoftware/IdentityServer/blob/main/src/IdentityServer/Validation/Default/SecretParser.cs#L43)

brockallen:
Thanks for the PR and suggestion. We've opened an issue to consider for 7.0. Thanks!

Issue #1111, opened 2022-12-12
Filter in Session Manager sample UI doesn't work


nikriaz:
**Which version of Duende IdentityServer are you using?** 6.2

Filter in the Session Manager sample UI doesn't work and cannot work. If the intention was to make an "universal filter" by any of three properties then logic in the `QuerySessionsAsync ` should be changed. Sorry for bothering you by so small issues. 

```
                UserSessions = await _sessionManagementService.QuerySessionsAsync(new SessionQuery
                {
                    ResultsToken = Token,
                    RequestPriorResults = Prev == "true",
                    DisplayName = Filter,
                    SessionId = Filter,
                    SubjectId = Filter,
                });

            query = query.Where(x =>
                (filter.SubjectId == null || x.SubjectId.Contains(filter.SubjectId)) &&
                (filter.SessionId == null || x.SessionId.Contains(filter.SessionId)) &&
                (filter.DisplayName == null || (x.DisplayName != null && x.DisplayName.Contains(filter.DisplayName) == true))
```

PR #1218, opened 2023-03-17
Fix server side session ui filtering

josephdecock:
Resolves https://github.com/DuendeSoftware/IdentityServer/issues/1111


brockallen:
@josephdecock we should chat about this one. i'm thinking the prior fix I made was actually wrong.

brockallen (approved):

PR #1220, opened 2023-03-20
Add Configuration projects to build

leastprivilege:


leastprivilege:
This should be it. I will look into the CodeQL configuration warning.

@josephdecock I noticed that that you produce console output during the build
https://github.com/DuendeSoftware/IdentityServer/actions/runs/4465530078/jobs/7842678375#step:6:146

Is that intentional?

josephdecock:
> @josephdecock I noticed that that you produce console output during the build https://github.com/DuendeSoftware/IdentityServer/actions/runs/4465530078/jobs/7842678375#step:6:146
> 
> Is that intentional?

This looks like a bug in the integration tests' host, but I'm not able to reproduce it locally.

The test framework is largely a copy of the test framework code from the BFF, but I'm using `WebApplicationBuilder` instead of `IHostBuilder` because I am creating the dcr endpoint using MapPost. This change meant I had to rearrange some of the service registration in the test hosts as well, and I suspect that the test double for logging isn't being registered properly anymore.

So far, I've tried a couple of different changes to try to address this, but since I can't repro locally, it is a bit tedious.

brockallen (approved):

PR #1233, opened 2023-03-30
add some extra logging to dpop validation

brockallen:


leastprivilege (approved):

src/IdentityServer/Validation/Default/DefaultDPoPProofValidator.cs
```diff
@@ -305,7 +305,7 @@ protected virtual async Task ValidatePayloadAsync(DPoPProofValidatonContext cont
         await ValidateReplayAsync(context, result);
         if (result.IsError)
         {
-            Logger.LogDebug("Detected replay of DPoP token");
+            result.ErrorDescription = "Detected replay of DPoP proof token.";
```
josephdecock, line 308: Does this get sent back to the attacker? If so, why?

brockallen: Yes -- but do you feel this info is a problem?

josephdecock: Well, it leaks information to the attacker that this DPOP implementation is tracking tokens for replay. That seems minor, but I guess that might prompt an attacker to try to perform some sort of memory exhaustion attack. That brings up a separate issue, which is that we should be defending against memory exhaustion by rejecting jtis that are unreasonably long.

brockallen: > Well, it leaks information to the attacker that this DPOP implementation is tracking tokens for replay.

Well, the code is on the internet :)

> which is that we should be defending against memory exhaustion by rejecting jtis that are unreasonably long

Good point. I thought we had done a size check somewhere. I'll look into this.

PR #1265, opened 2023-05-02
Add post configuration of identity server options

okhara:
**Redirecting back to client application after logging out from the external identity provider**

User is not redirected back to the client application from external identity provider, when identity server is configured for asp.net identity.

[Link to the Issue](https://github.com/DuendeSoftware/IdentityServer/issues/1404)

After some debugging of the duende and Microsoft sources, I've discovered that the reason for the redirection problem is the absence of id token. SignOutScheme which is used to retrieve id token from the user is set to default value, but default authentication scheme after configuring identity server for asp.net identity is changed to Application.Identity.

### DIFF
--- a/src/AspNetIdentity/PostConfigureIdentityServerOptions.cs
+++ b/src/AspNetIdentity/PostConfigureIdentityServerOptions.cs
@@ -1,0 +1,31 @@
+using Duende.IdentityServer.Configuration;
+using Microsoft.Extensions.Options;
+
+namespace Duende.IdentityServer.AspNetIdentity;
+
+/// <summary>
+/// Identity server options configuration
+/// </summary>
+public class PostConfigureIdentityServerOptions : IPostConfigureOptions<IdentityServerOptions>
+{
+    private readonly IOptions<Microsoft.AspNetCore.Authentication.AuthenticationOptions> _authOptions;
+
+    /// <summary>
+    /// ctor
+    /// </summary>
+    /// <param name="authOptions"></param>
+    public PostConfigureIdentityServerOptions(IOptions<Microsoft.AspNetCore.Authentication.AuthenticationOptions> authOptions)
+    {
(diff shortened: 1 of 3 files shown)


src/AspNetIdentity/PostConfigureIdentityServerOptions.cs
```diff
@@ -0,0 +1,40 @@
...
+
+    /// <inheritdoc/>
+    public void PostConfigure(string name, IdentityServerOptions options)
+    {
+        // If we are using ASP.NET Identity and the dynamic providers don't have a
+        // sign out scheme set, then we need the dynamic providers to use ASP.NET
+        // Identity's cookie at sign out time. If the sign out scheme is explicitly
+        // set, then we don't override that though.
+
+        if (DefaultAuthSchemeIsAspNetIdentity() &&
```
josephdecock, line 31: This logic has been refactored a bit to make the intentions clearer

src/IdentityServer/Configuration/DependencyInjection/Options/DynamicProviderOptions.cs
```diff
@@ -31,7 +31,24 @@ public class DynamicProviderOptions
...
+            _signOutScheme = value;
+        }
+    }
+    
+    private string? _signOutScheme;
+
+    /// <summary>
+    /// Gets a value indicating if the SignOutScheme was set explicitly, either by application logic or by options binding.
+    /// </summary>
+    public bool SignOutSchemeSetExplicitly { get => _signOutScheme != null; }
```
josephdecock, line 51: The options class now understands if the sign out scheme has actually been set. We use that in our configuration so that we don't override a user who is trying to set the scheme explicitly.

josephdecock:
@brockallen I've made some changes to this PR, would you take a look and review please?

brockallen (approved):

Issue #1266, opened 2023-05-03
Consider flag to disable DPoP


brockallen:
https://github.com/DuendeSoftware/docs.duendesoftware.com/pull/265#discussion_r1184354552

leastprivilege:
I think I would prefer to fail if a proof token is received and not the right license...

brockallen:
Something to consider -- if there is no license, should we still emit a log entry indicating that this is a licensed feature? I'm thinking the scenarios where a customer is evaluating or in dev mode and don't know it's a higher level feature.

josephdecock:
> not the right license

This is distinct from no license at all? So non prod environments would be usable?

brockallen:
Well, that's still open for discussion. If there's no license at all I'd say it should work, but I proposed we log that you're using an Enterprise level feature so you'd know you can't just then buy Starter, for example.

leastprivilege:
agreed

PR #1267, opened 2023-05-05
don't accept dpop request is license does not allow

brockallen:
Closes: https://github.com/DuendeSoftware/IdentityServer/issues/1266

### DIFF
--- a/src/IdentityServer/Validation/Default/LicenseValidator.cs
+++ b/src/IdentityServer/Validation/Default/LicenseValidator.cs
@@ -176,12 +176,19 @@ internal class LicenseValidator
         }
     }
 
-    public static bool ValidateDPoP()
+    public static bool CanUseDPoP()
     {
-        if (_license != null && !_license.DPoPFeature)
+        if (_license != null)
         {
-            _errorLog.Invoke("A request was made using DPoP. Your license for Duende IdentityServer does not include the DPoP feature.", Array.Empty<object>());
-            return false;
+            if (!_license.DPoPFeature)
+            {
+                _errorLog.Invoke("A request was made using DPoP. Your license for Duende IdentityServer does not include the DPoP feature.", Array.Empty<object>());
+                return false;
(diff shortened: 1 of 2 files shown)


brockallen:
> I was thinking that license error was only triggering log errors/warnings and cannot impact production. 

Yes, for license limits that are exceeded like client count or license expiration which are more understandable as an oversight, and it's those scenarios that we did not want to disable the operations as a result

> Is this feature the only one that will reject a request ?

Yes, it would be the first. But if your license didn't allow in the first place, then you should not have been using in the first place and this will prevent you from accidentally using it. We're also thinking of adding logs to let you know in dev mode that this feature requires a certain license type.

Thoughts?

g7ed6e:
> > I was thinking that license error was only triggering log errors/warnings and cannot impact production. 
> 
> 
> 
> Yes, for license limits that are exceeded like client count or license expiration which are more understandable as an oversight, and it's those scenarios that we did not want to disable the operations as a result
> 
> 
> 
> > Is this feature the only one that will reject a request ?
> 
> 
> 
> Yes, it would be the first. But this is a feature you don't tend to accidentally use, and so if your license didn't allow in the first place, then you should not have been using in the first place so this will prevent you from using it. We're also thinking of adding logs to let you know in dev mode that this feature requires a certain license type.
> 
> 
> 
> Thoughts?

@brockallen my concern is about a licence expiration that would force to patch an application outside of its own roadmap. If license may break things it could be interesting to be able to hot reload it but I think that's not possible as of now to hot reload IdentityServerOptions..

brockallen:
> @brockallen my concern is about a licence expiration that would force to patch an application outside of its own roadmap. If license may break things it could be interesting to be able to hot reload it but I think that's not possible as of now to hot reload IdentityServerOptions..

License expiration would not then suddenly start to prevent DPoP if the license normally would allow it. That is not the intent with this check.

PR #1285, opened 2023-05-18
Filter out expected ef migration exception

josephdecock:
Filter out exceptions that we expect to see during migrations in the asp.net identity and EF hosts. We already do this in the quickstart tutorial samples.

Should help folks who encounter problems like this: https://github.com/DuendeSoftware/Support/issues/678

### DIFF
--- a/hosts/AspNetIdentity/Program.cs
+++ b/hosts/AspNetIdentity/Program.cs
@@ -34,7 +34,12 @@ try
 
     app.Run();
 }
-catch (Exception ex)
+catch (Exception ex) when (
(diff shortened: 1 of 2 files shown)


brockallen:
.NET 8 is coming soon, and they expose that exception type, right? Maybe this should use that new one (once we update our dependencies)?

brockallen:
Or was the OP using this host as the starter/sample?

josephdecock:
I was thinking we'd include this in the templates for now, and switch to StopTheHostException for the next release.

brockallen:
This is in the template already, and we're going to re-do for .NET 8 soon anyway.

PR #1290, opened 2023-05-19
Remove "tools: latest" from codeql

josephdecock:


brockallen:
this is removing AspId host so build doesn't fail

brockallen (approved):

PR #1297, opened 2023-05-23
Joe/session issued time

josephdecock:


src/IdentityServer/Extensions/AuthenticationTicketExtensions.cs
```diff
@@ -156,6 +156,10 @@ public static AuthenticationTicket Deserialize(this ServerSideSession session, I
                 properties.ExpiresUtc = null;
             }
 
+            // similar to the expires update above, this is needed so that the ticket will have the 
+            // right issued timestamp when the cookie handler performs its sliding logic
+            properties.IssuedUtc = new DateTimeOffset(session.Renewed, TimeSpan.Zero);
+
```
josephdecock, line 162: ServerSideTicketStore's query etc methods may need to change as well.

brockallen (approved):

Issue #1273, opened 2023-05-10
Break license validation into partial classes for easier sharing across projects


brockallen:
IS, BF, ConfigApi

PR #1299, opened 2023-05-24
Rework/Reorg license key validation

brockallen:
Closes: #1273


brockallen:
Do we want diff or same names for loggers?

brockallen:
Ok, good enough for now. We have time to refine as we progress.

josephdecock (approved):

Issue #1268, opened 2023-05-08
Handling of licensed features going forward


leastprivilege:
to expand on https://github.com/DuendeSoftware/IdentityServer/issues/1266

Maybe we should do this:

* no license: always emit warning when a feature is used that requires an Edition > Starter
* license present: fail if feature is used that is not licensed

We probably need to make an exception for using multiple issuers because that often happens by accident - IOW warn only.

brockallen:
For 6.3 or 7.0?

brockallen:
Submitted a PR to do the logging piece. I anticipate feedback that we're now spamming the logs.

PR #1305, opened 2023-05-30
emit warning logs when no license present and licensed features are being used

brockallen:
Closes: https://github.com/DuendeSoftware/IdentityServer/issues/1268


brockallen:
I think I'm happy with this now @josephdecock

josephdecock:
Do we need to mention this in the docs? https://docs.duendesoftware.com/identityserver/v6/fundamentals/license_key/

brockallen:
@josephdecock Onyl if what's in the docs there is no longer correct.

PR #1319, opened 2023-06-02
Make License available in DI

brockallen:


src/IdentityServer/Licensing/LicenseValidator.cs
```diff
@@ -15,41 +15,48 @@
...
+            _copy.Initalize(License.Claims.Clone());
+        }
+        return _copy;
+    }
 
-    static void Initalize(ILoggerFactory loggerFactory, string productName, string key, bool isDevelopment = false)
+    protected void Initalize(ILoggerFactory loggerFactory, string productName, string key)
     {
-        _logger = loggerFactory.CreateLogger($"Duende.{productName}.License");
+        //if (Logger != null) throw new InvalidOperationException("LicenseValidator already initialized.");
```
josephdecock, line 49: Do we want this commented out?

brockallen: I removed it because I don't think it hurts to re-run it and I thought it might be possible that people might call this more than once for some reason.

josephdecock (approved):

Issue #1254, opened 2023-04-25
Consider an enhanced profile service when using server-side sessions


brockallen:
e.g. so that the user info endpoint can draw from a user's session claims (stored in the DB).

and possibly the approach here is loading the Subject from the server-side session table before calling into the profile service.

brockallen:
related: https://github.com/DuendeSoftware/IdentityServer/issues/884

Issue #884, opened 2022-05-11
Research preserving the user's session at userinfo/introspection


brockallen:
This would mean adding the original Subject to the grant (beyond the user's claims in the AT).

https://github.com/DuendeSoftware/Support/issues/56

PR #1327, opened 2023-06-06
When using server-side sessions, load subject from server side session store if in use

brockallen:
This PR changes how the introspection endpoint works when using server side sessions. The Subject now passed to the profile service will be the ClaimsPrincipal from the server-side session store, rather than a "minimal" ClaimsPrincipal synthesized from the claims in the access token. This allows the userinfo endpoint claims generation to now access claims that are only determined as part of the user's session (in addition to anything previously accessible from the user profile store).

Closes: https://github.com/DuendeSoftware/IdentityServer/issues/1254
Closes: https://github.com/DuendeSoftware/IdentityServer/issues/884


josephdecock (approved):
Should we add a test? Maybe not, could be a lot of setup for not much value.

brockallen:
> Should we add a test? Maybe not, could be a lot of setup for not much value.

Actually I forgot that our existing tests already have code coverage that executes this new code in this PR. That's why I didn't feel we needed any additional tests.

Issue #1129, opened 2023-01-13
Consider result generator in DI for our result objects


brockallen:
As we discussed for the DCR architecture and how MVC decouples the response object from the response rendering.

PR #1330, opened 2023-06-07
Support response generators for the endpoint results

brockallen:
This PR introduces the `IEndpointResultGenerator<T>` type which is designed to render the `IEndpointResult` objects that are returned from the `IEndpointHandler` in our endpoint routing architecture.

Previously the `IEndpointResult` would render themselves, but was difficult for customers to override or change how they did this. Now our built in `IEndpointResult` classes use a new base class `EndpointResult` which will resolve a `IEndpointResultGenerator<T>` from DI to do the rendering. This allows customers to register their own `IEndpointResultGenerator<T>` and can change how they are rendered to the HTTP response.

This PR is also designed to be backwards compatible with the prior approach, such that any custom `IEndpointResult` or `IEndpointHandler` should still work the way they used to. 

Closes: https://github.com/DuendeSoftware/IdentityServer/issues/1129


brockallen:
This PR got borked by rebasing. Will fix.

Issue #845, opened 2022-03-31
Consider cleaning up InteractionResponse processing


brockallen:
Perhaps adding enum for the types of response: Login, Consent, Error, etc.?
Generalize redirect processing?
Allow for more flexible types?

related: https://github.com/DuendeSoftware/IdentityServer/issues/432

brockallen:
Related: when an invalid prompt param is passed, we ignore. the newer spec says we should treat as error, but this would be a behavioral change, thus we'd want to hold off until 7.0.0.

> If the OpenID Provider receives a prompt value that it does not support (not declared in the prompt_values_supported metadata field) the OP SHOULD respond with an HTTP 400 (Bad Request) status code and an error value of invalid_request. It is RECOMMENDED that the OP return an error_description value identifying the invalid parameter value.

PR #1331, opened 2023-06-07
return error for unknown prompt values on authorize endpoint

brockallen:
Fixes: https://github.com/DuendeSoftware/IdentityServer/issues/845#issuecomment-1405377531


brockallen:
Not sure how we document this as of now.

josephdecock (approved):

PR #1341, opened 2023-06-22
Use new clock abstraction

brockallen:
Given that the old ISystemClock has only second-level granularity, and that it is deprecated, this PR introduces a new clock abstraction. The default implementation is based on the new TimeProvider in .NET 8, but a legacy implementation is provided for those who still want the older ISystemClock behavior.

Closes: https://github.com/DuendeSoftware/IdentityServer/issues/949


brockallen:
pinging @kimayoi72 @BrianAlk fore review/feedback, since you had opened issues in the past related to this.

josephdecock (approved):

PR #1344, opened 2023-06-26
Fixes nullability annotation on IReturnUrlParser.ParseAsync

### LINKED ISSUE #1343: OidIReturnUrlParser nullability does not conform to IReturnUrlParser.ParseAsync

Latest

**Describe the bug**

`OidIReturnUrlParser.ParseAsync` returns `null` when the `returnUrl` is not valid, but the interface (`IReturnUrlParser`) and its implementation state that the return value is non-nullable.

josephdecock:
Closes #1343.


josephdecock:
TODO - remove from implementation (only do in interface)

brockallen (approved):

PR #1346, opened 2023-06-26
Fix log messages in backchannel user validator

### LINKED ISSUE #1345: BackchannelAuthenticationRequestValidator inserts an unwanted string format item when calling LogError

6.2.3

.NET 6

**Describe the bug**

__BackchannelAuthenticationRequestValidator__ calls __LogError__ with a string format item which results in an exception when __LogWithRequestDetails__ is called. 

A permalink directly to the line that causes the error:
https://github.com/DuendeSoftware/IdentityServer/blob/d933a168074d7248226241acd1b96d1910baf8bd/src/IdentityServer/Validation/Default/BackchannelAuthenticationRequestValidator.cs#LL410C9-L410C9

**To Reproduce**

Implement __IBackchannelAuthenticationUserValidator__ and make a ciba request that results in any __Error__ value that is not one of the below:
* OidcConstants.BackchannelAuthenticationRequestErrors.AccessDenied
* OidcConstants.BackchannelAuthenticationRequestErrors.ExpiredLoginHintToken
* OidcConstants.BackchannelAuthenticationRequestErrors.UnknownUserId
* OidcConstants.BackchannelAuthenticationRequestErrors.MissingUserCode
* OidcConstants.BackchannelAuthenticationRequestErrors.InvalidUserCode
* OidcConstants.BackchannelAuthenticationRequestErrors.InvalidBindingMessage

**Actual behavior**

__LogError__ calls __LogWithRequestDetails__ with the string format item "{error}" within the message which will result in the following error being logged:
```
Error logging An error occurred while writing to logger(s). (Index (zero based) must be greater than or equal to zero and less than the size of the argument list.), request details: 
```

**Expected behavior**

LogError logs the error with the Error value from __BackchannelAuthenticationUserValidationResult.Error__ :
```
Unexpected error from IBackchannelAuthenticationUserValidator: <The error returned from the implementation of IBackchannelAuthenticationUserValidator.ValidateRequestAsync> 
```

**Additional context**

This arises when we have an implementation of __IBackchannelAuthenticationUserValidator__ where __ValidateRequestAsync__ populates the field __BackchannelAuthenticationUserValidationResult.Error__ with a value that is not one of the below:
* OidcConstants.BackchannelAuthenticationRequestErrors.AccessDenied
* OidcConstants.BackchannelAuthenticationRequestErrors.ExpiredLoginHintToken
* OidcConstants.BackchannelAuthenticationRequestErrors.UnknownUserId
* OidcConstants.BackchannelAuthenticationRequestErrors.MissingUserCode
* OidcConstants.BackchannelAuthenticationRequestErrors.InvalidUserCode
* OidcConstants.BackchannelAuthenticationRequestErrors.InvalidBindingMessage

This does not cause any actual errors or unexpected behavior in the ciba flow itself, but the error logs do not log the expected errors because of the additional string format item.

josephdecock:
This fixes #1345, which happened because the passed in log message contained a placeholder, and LogWithRequestDetails also added a placeholder for the values, resulting in too many placeholders.

The fix is to remove the extra placeholder from LogWithRequestDetails and review all calls to it to ensure the passed log message always contains placeholders for the values.

### DIFF
--- a/src/IdentityServer/Validation/Default/BackchannelAuthenticationRequestValidator.cs
+++ b/src/IdentityServer/Validation/Default/BackchannelAuthenticationRequestValidator.cs
@@ -447,7 +447,7 @@ internal class BackchannelAuthenticationRequestValidator : IBackchannelAuthentic
             var payloadClientId = jwtRequestValidationResult.Payload.SingleOrDefault(x => x.Type == JwtClaimTypes.ClientId)?.Value;
             if (payloadClientId.IsPresent() && _validatedRequest.Client.ClientId != payloadClientId)
             {
-                LogError("client_id found in the JWT request object does not match client_id used to authenticate", new { invalidClientId = payloadClientId, clientId = _validatedRequest.Client.ClientId });
+                LogError("client_id found in the JWT request object does not match client_id used to authenticate, {@values}", new { invalidClientId = payloadClientId, clientId = _validatedRequest.Client.ClientId });
                 return (false, Invalid(OidcConstants.AuthorizeErrors.InvalidRequestObject, "Invalid client_id in JWT request"));
             }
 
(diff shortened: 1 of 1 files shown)


josephdecock:
TODO - Decide if and where to patch this, then rebase onto correct release branch.

josephdecock:
@brockallen - this is ready for review, please.

brockallen (approved):

PR #1361, opened 2023-07-07
Adds Custom property to BackchannelAuthenticationResponse

gislikonrad:
### The problem
We have an authenticator app that is similar to other authenticator apps where when out-of-band authentication is invoked, a code is displayed in the invoking client that needs to be verified in the authenticator app. We need a way to send the client this verification code during CIBA authentication.

### The spec
According to section 3 in the CIBA spec _it allows for some extensibility - profiles can be created that define additional parameters to be sent to, **and received from**, the Backchannel Authentication Endpoint._ 

### The proposed solution
What I did was to add the preliminary `BackchannelAuthenticationResponse` to the `BackchannelUserLoginRequest` so that it can be mutated in the `IBackchannelAuthenticationUserNotificationService`. A new dictionary property was added to the `BackchannelAuthenticationResponse` which is similar to the one in `TokenResponse`. This dictionary is then added to the DTO as JsonExtensionData to be serialized.

### More
I considered adding the raw request to the `BackchannelUserLoginRequest` aswell, to be able to respond to extra request parameters, but I might add that to a separate PR.

---

~There is no issue reported for this, but I felt that explaining my issue would be easier in code. We are running v6.2.3 of Duende.IdentityServer.~
[An issue was created](https://github.com/DuendeSoftware/IdentityServer/issues/1405) for this feature request.


src/IdentityServer/ResponseHandling/Default/BackchannelAuthenticationResponseGenerator.cs
```diff
@@ -111,6 +111,7 @@ public virtual async Task<BackchannelAuthenticationResponse> ProcessAsync(Backch
             AuthenticationContextReferenceClasses = validationResult.ValidatedRequest.AuthenticationContextReferenceClasses,
             Tenant = validationResult.ValidatedRequest.Tenant,
             IdP = validationResult.ValidatedRequest.IdP,
+            Response = response
```
josephdecock, line 114: Why are we sending the response to the user login service? If we want to return custom values in the response from the CIBA endpoint, that should be passed above (line 101).

josephdecock (commented):
I think maybe we need to add a new custom validator abstraction, similar to what is done on the token endpoint. Then users could provide an implementation that would build up the Custom dictionary.

src/IdentityServer/Models/BackchannelUserLoginRequest.cs
```diff
@@ -59,4 +60,9 @@ public class BackchannelUserLoginRequest
     /// Gets or sets the validated resources.
     /// </summary>
     public ResourceValidationResult ValidatedResources { get; set; } = default!;
+
+    /// <summary>
+    /// The response that will be sent to the client
+    /// </summary>
+    public BackchannelAuthenticationResponse Response { get; set; } = default!;
```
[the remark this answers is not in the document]
gislikonrad: Passing just the Custom dictionary would be sufficient.

src/IdentityServer/ResponseHandling/Default/BackchannelAuthenticationResponseGenerator.cs
gislikonrad: If the custom dictionary would just be added on line 101, then the user login service couldn't add anything to the response, i.e a code that should be shown on a page and entered in the backchannel client.

gislikonrad:
A custom validator abstraction would also work.

josephdecock:
Closing in favor of new PR 1497 (this branch was rebased, and the custom validator was added).

PR #1363, opened 2023-07-07
Remove raw exception from unhandled exception event

### LINKED ISSUE #1362: UnhandledExceptionEvent.ToString() Throws NotSupportedException
6.3.2
NET 6
**Describe the bug**

1. System.Text.Json can't serialize exceptions. https://github.com/dotnet/runtime/issues/43026. It will refuse to serialize the contained IntPtr
2. UnhandledExceptionEvent contains an exception https://github.com/DuendeSoftware/IdentityServer/blob/main/src/IdentityServer/Events/UnhandledExceptionEvent.cs#L44
3. Event.ToString() calls LogSerializer.Serialize(this) which internally uses System.Text.Json https://github.com/DuendeSoftware/IdentityServer/blob/main/src/IdentityServer/Events/Infrastructure/Event.cs#L145

**To Reproduce**

````
new UnhandledExceptionEvent(new Exception()).ToString();
````

should do it, although I haven't exactly tried, because in my case it happens during the OIDC flow (likely because I haven't migrated the database yet).

josephdecock:
This would fix #1362. Seems a bit heavy-handed to just completely remove the exception, but it can't be serialized anyway so I don't think we're losing much.

### DIFF
--- a/src/IdentityServer/Events/UnhandledExceptionEvent.cs
+++ b/src/IdentityServer/Events/UnhandledExceptionEvent.cs
@@ -23,7 +23,6 @@ public class UnhandledExceptionEvent : Event
             EventIds.UnhandledException,
             ex.Message)
     {
-        Exception = ex;
         Details = ex.ToString();
     }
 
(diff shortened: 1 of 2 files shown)


brockallen:
Given that we already capture the exception in the `Details` property then I think it's ok to remove it.

PR #1366, opened 2023-07-11
Introducing Refresh Token Leakage Detection Event

amiru3f:
Hello there,

I am excited to contribute to Duende Identity Server by proposing some feature improvements. These enhancements aim to meet our specific requirements while also aligning with the OAuth 2.0 standard.

Problem:
One crucial area that requires attention is Refresh Token Protection. According to the OAuth standard (as mentioned in [this link](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics-13#section-4.12)), the leakage of a refresh token can have significant consequences, depending on the scale of the problem. However, Duende currently lacks a module or specific event dedicated to handling this issue or at least detecting that

Proposed Solution:
To address this gap, I suggest raising an event in case of refresh token misuse or double spend. This feature would focus on detection not prevention. By raising an event when double spending of a refresh token occurs, the developers gain flexibility in their codebase. Additionally, they can capture valuable information such as the IP address, communication channel, and subject concurrently with the event. The decision on how to handle this event can be delegated to the application, as it falls under our application's responsibility rather than Duende's.

Further Considerations:
I would like to clarify that double spending, in this context, does not pertain to whether or not we accept double consumed tokens. We have the option to either accept or reject multiple consumed refresh tokens, and we can keep a count of them for use in other decision-making processes.

I believe that by introducing this Refresh Token Leakage Detection and Prevention feature, developers can greatly enhance the security and flexibility of their codes and Duende Identity Server, while adhering to established standards.

Thank you for considering this contribution.

### DIFF
--- a/src/IdentityServer/Configuration/DependencyInjection/Options/EventsOptions.cs
+++ b/src/IdentityServer/Configuration/DependencyInjection/Options/EventsOptions.cs
@@ -41,4 +41,12 @@ public class EventsOptions
     ///   <c>true</c> if error events should be raised; otherwise, <c>false</c>.
     /// </value>
     public bool RaiseErrorEvents { get; set; } = false;
+
+    /// <summary>
+    /// Gets or sets a value indicating whether to raise error events.
+    /// </summary>
+    /// <value>
+    ///   <c>true</c> if error events should be raised; otherwise, <c>false</c>.
+    /// </value>
+    public bool RaiseWarningEvents { get; set; } = false;
 }
--- a/src/IdentityServer/Events/Infrastructure/EventIds.cs
+++ b/src/IdentityServer/Events/Infrastructure/EventIds.cs
@@ -18,7 +18,7 @@ public static class EventIds
 
     public const int ClientAuthenticationSuccess = AuthenticationEventsStart + 10;
     public const int ClientAuthenticationFailure = AuthenticationEventsStart + 11;
-        
+
     public const int ApiAuthenticationSuccess = AuthenticationEventsStart + 20;
     public const int ApiAuthenticationFailure = AuthenticationEventsStart + 21;
 
@@ -34,7 +34,7 @@ public static class EventIds
 
     public const int TokenIntrospectionSuccess = TokenEventsStart + 20;
     public const int TokenIntrospectionFailure = TokenEventsStart + 21;
-        
+
     //////////////////////////////////////////////////////
     /// Error related events
     //////////////////////////////////////////////////////
@@ -68,4 +68,9 @@ public static class EventIds
 
     public const int BackchannelAuthenticationSuccess = BackchannelAuthenticationEventsStart + 0;
     public const int BackchannelAuthenticationFailure = BackchannelAuthenticationEventsStart + 1;
+
+    //////////////////////////////////////////////////////
+    /// Token usage related events
+    //////////////////////////////////////////////////////
+    public const int TokenMissuseEvent = 7000;
 }
--- a/src/IdentityServer/Events/Infrastructure/EventType.cs
+++ b/src/IdentityServer/Events/Infrastructure/EventType.cs
@@ -27,5 +27,10 @@ public enum EventTypes
     /// <summary>
     /// Error event
     /// </summary>
-    Error = 4
+    Error = 4,
+
+    /// <summary>
+    /// Warning event
+    /// </summary>
+    Warning = 5
 }
--- a/src/IdentityServer/Events/RefreshTokenDoubleSpentEvent.cs
+++ b/src/IdentityServer/Events/RefreshTokenDoubleSpentEvent.cs
@@ -1,0 +1,56 @@
+// Copyright (c) Duende Software. All rights reserved.
+// See LICENSE in the project root for license information.
+
+
+using Duende.IdentityServer.Models;
+
+namespace Duende.IdentityServer.Events;
+
+/// <summary>
+/// Event for double spent refresh token
+/// </summary>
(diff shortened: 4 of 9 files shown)


brockallen:
Hi @amiru3f -- Just a quick question: Is this something you could achieve in your own IdentityServer implementation without extending our core?

amiru3f:
Hi @brockallen 
Thank you for taking the time to review my pull request

Actually yes, I will capture the `RefreshTokenDoubleSpentEvent` and there are several useful parameters in the consumed refresh token. (SessionId, SubjectId, ClientId etc.)

By capturing the `TokenIssuedSuccessEvent` event, I store a `chain` of issued (and rotated) refresh tokens (with specific expiration time for each element). 

`[sessionId]` =>  [ref_token1] -> [ref_token2] -> [ref_token3] -> ...  -> [last_ref_token]
OR
`[sub + client]` =>  [ref_token1] -> [ref_token2] -> [ref_token3] -> ...  -> [last_ref_token]

I will find the latest element in my stored chain (which is the `refresh_token` issued by the first consumption of double-spent-refresh-token)  and call the revoke with the key.

I'm also thrilled to submit a pull request after my own successful implementation, to have it supported by Duende in the box

josephdecock:
@amiru3f, we're so glad for your enthusiasm with IdentityServer! 

Our plans for observability are actually to de-emphasize Events in the future. Our thinking is that OpenTelemetry offers a lot of advantages, so we want to emphasize and expand our support for OTel, rather than our events infrastructure. The biggest advantage of OTel is probably that by following these standards, the signals that IdentityServer emits can be consumed by other tools. For example, right now, we support traces, and you can consume the trace data in tools like honeycomb.io (or any other OTel service) to get visualizations of your traces without any additional configuration or setup. In the future we plan to support metrics and possibly OTel logs, which hopefully will have similar tooling benefits.

Since we're moving in this direction of emphasizing the OTel support, we don't want to expand our events infrastructure with more events - as long as you're able to accomplish what you need without making changes to the core of IdentityServer.

IdentityServer's events aren't going away - that would be a big breaking change for existing users of events!  Continuing to use events and extending them with your own custom events is a fine thing to do, and supported in IdentityServer. But as I say, we're in the process of trying to reduce how much we use them, expand our support of OTel, and encourage folks to use that instead.

So, I'm sorry to say that we aren't going to merge this PR. Again, we love to see your enthusiasm for our framework, and we thank you, even though we're not merging it.

amiru3f:
@josephdecock 

Thank you for your thorough response. Honestly, I had also been wondering why Duende is not utilizing Diagnostics/OTEL for its events/metrics. I attempted to capture that specific event and push another event to IDiagnosticsSource in my application, intending to observe it elsewhere and implement my own logic.

On another note, without this event, I am left with the option to extend `DefaultRefreshTokenService` and duplicate all of the code from the `ValidateRefreshTokenAsync` method in my application, just to manually trigger that event. Consequently, with each change to this class in future versions, I would have to update my code accordingly.

In the meantime, I am eager to contribute to the Duende project and add the feature of Leakage detection, if you find it suitable.

josephdecock:
We actually have a different virtual method in the refresh token service that you can override to emit your event without needing to replace the entire validation method.

It's called `AcceptConsumedTokenAsync` and it is called when identity server sees a  refresh token with rotation semantics more than once. You return true or false from it to decide if you want to allow the refresh token to be used. The default implementation just returns false every time, but your override could add more sophisticated logic and _emit your custom event._

More details are in the docs here
https://docs.duendesoftware.com/identityserver/v6/tokens/refresh/#replay-detection.

amiru3f:
Thanks man, it works

amiru3f:
Hey there,


@josephdecock  With respect to the conversation upper, I've completely implemented to `refresh token leakage detection + blocking the whole session`, which it triggers right after double consuming a `refresh token` (with some criteria, configurations and detection strategies in our code)

The criteria might be something like:
* Sliding and absolute lifetime of the refresh token which is configured at `Client` entity
* IP of the user
* etc

Since the process starts right after execution of `AcceptConsumedTokenAsync` and acts like blocking the whole chain of the session **ultimately**, I would again need to check the `Client` configurations. Although the object was fetched once in the TokenRequest, I can't reuse it here cuz it's not passed in the `AcceptConsumedTokenAsync` args

Does it make sense to have it like the one is passed in 
```
ValidateRefreshTokenAsync(string tokenHandle, Client client)
```

Sth like changing
```
AcceptConsumedTokenAsync(RefreshToken refreshToken) 
```
to
```
AcceptConsumedTokenAsync(RefreshToken refreshToken, Client client = null)
```
So that I wouldn't need to reuse the`IClientStore` for repeating the client fetch

amiru3f:
Actually I also made it that way by wrapping `FindClientByIdAsync` inside a memory cache to prevent db/api calls.
I was thinking about the complexities come in place for handling the cache invalidation and also fetching the client from store like:

``` c#
protected override async Task<bool> AcceptConsumedTokenAsync(RefreshToken refreshToken)
{
    if (await IsThisCandidateForBlocking(refreshToken))
    {
        await _destroyRefreshTokenChainCommandStore.Add(new(refreshToken.SubjectId, refreshToken.ClientId, refreshToken.SessionId));
    }

    return false;
}

private async Task<bool> IsThisCandidateForBlocking(RefreshToken refreshToken)
{
    var client = await _clientStore.FindClientByIdAsync(refreshToken.ClientId);
    return if client has some specific limits;
}
```

Anyways, you are right regarding the breaking change.
Thanks @josephdecock  for your reply.

PR #1375, opened 2023-07-19
Fix KeyManagementOptions binding

### LINKED ISSUE #1374: SigningAlgorithms ignored from appSettings.json
I am trying to configure, via appSettings.json, the SigningAlgorithms allowed.

As SigningAlgorithms is defined as
`public IEnumerable<SigningAlgorithmOptions> SigningAlgorithms { get; set; }`
with IEnumerable there is no .Add method therefore using the Options it cannot add the values I define in.  I found that information in [dotnet issue 63083](https://github.com/dotnet/runtime/issues/63083)

with this in my appSettings.json file, the values I define in SigningAlgorithms are ignored and result in only RS256
```
  "IdentityServerOptions": {
    "KeyManagement": {
      "Enabled": true,
      "SigningAlgorithms": [
        {
          "Name": "RS512",
          "UseX509Certificate": true
        },
        {
          "Name": "RS384"
        }
      ]
    }
  }
```

Am I doing something wrong?  Or is not expected to override the signing algorithms?

Thanks,

josephdecock:
Allow KeyManagementOptions to be bound from appsettings.json or
other config sources

Fixes #1374

### DIFF
--- a/src/IdentityServer/Configuration/DependencyInjection/Options/KeyManagementOptions.cs
+++ b/src/IdentityServer/Configuration/DependencyInjection/Options/KeyManagementOptions.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
 using static Duende.IdentityServer.IdentityServerConstants;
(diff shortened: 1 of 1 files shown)


josephdecock:
Technically this is a breaking change, as the interface changes from IEnumerable to ICollection. But to actually cause breakage, someone would have to be configuring key management options in code using a type that was an IEnumerable but wasn't an ICollection.

brockallen (approved):

PR #1377, opened 2023-07-19
Map entities to/from models by hand 

josephdecock:


josephdecock:
@brockallen, this is ready now

brockallen (approved):

PR #1385, opened 2023-07-27
Copyright Headers cleanup

josephdecock:
- Script to make all files have consistent copyright headers
- Remove apache license, copyright to brock, dom from 3 files

### DIFF
--- a/src/Storage/Models/SerializedKey.cs
+++ b/src/Storage/Models/SerializedKey.cs
@@ -1,3 +1,7 @@
+// Copyright (c) Duende Software. All rights reserved.
+// See LICENSE in the project root for license information.
+
+
 #nullable enable
 
 using System;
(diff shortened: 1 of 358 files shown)


src/Storage/Stores/IIdentityProviderStore.cs
```diff
@@ -1,5 +1,6 @@
-// Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
```
josephdecock, line 1: This is the most notable change.

src/Storage/Stores/ISigningKeyStore.cs
```diff
@@ -1,5 +1,6 @@
-// Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
```
josephdecock, line 1: Again, most notable change

src/Storage/Models/SerializedKey.cs
```diff
@@ -1,5 +1,6 @@
-// Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
```
josephdecock, line 1: 3rd and final instance

brockallen (approved):

PR #1395, opened 2023-08-18
Update xmldoc for InMemoryCorsPolicyService.cs

josephdecock:
Related to https://github.com/DuendeSoftware/Support/issues/834


brockallen:
Is there a comment in there somewhere that also informs that it's expected to only be used when using in-mem clients?

josephdecock:
Good point. I've updated to call out that you should be using this with in memory clients.

PR #1401, opened 2023-08-30
Updates for .NET8 Preview 7

leastprivilege:
...maybe we want to wait for preview 8. As expected the JWT handler has some regressions (e.g. "sub claim is missing").


brockallen:
Also, @josephdecock, take note of the warnings about obsolete APIs that are being used. We should find the new/replacement APIs and use those now instead.

leastprivilege:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/releases/tag/7.0.0

josephdecock:
All the warnings are now resolved as well.

- There was a warning about DbSet.Add vs DbSet.AddAsync. We should not change to AddAsync here.
  > This method is async only to allow special value generators, such as the one used by 'Microsoft.EntityFrameworkCore.Metadata.SqlServerValueGenerationStrategy.SequenceHiLo', to access the database asynchronously. For all other cases the non async method should be used.
- There was a warning about setting http headers by calling Add on the dictionary of headers. This throws if you've already set the header. In our case, we were setting the www-authenticate header this way. It actually would be an error if the www-authenticate header was set twice, so I check for that explicitly now.

brockallen (approved):

PR #1414, opened 2023-09-11
Fix optional license dependency

josephdecock:
Minor fix for an optional dependency


brockallen:
How does this manifest?

josephdecock:
If you have no license, then the factory method in DI for IdentityServerLicense returns null. Then when you try to go this page, DI throws an exception.

brockallen:
Ah yes, I think I recall we discussed this. Thanks.

PR #1423, opened 2023-09-20
Return invalid_request when no response_type is provided

user1336:
If you agree on https://github.com/DuendeSoftware/Support/issues/907.


AndersAbel (approved):

user1336:
Thanks @AndersAbel :shipit:

josephdecock:
@brockallen do you think this should be marked as a breaking change since it is a behavioral change?

brockallen:
IMO, it's still an error response the only diff is the error code. And if this is more correct WRT to the spec, then it's more of a bug fix.

Practically, I don't see a lot of clients looking at specific error codes. It's either success or error. The error code is more for debugging.

PR #1424, opened 2023-09-21
Add support for PAR

josephdecock:


src/Storage/Models/Client.cs
```diff
@@ -216,14 +216,28 @@ public ICollection<string> AllowedGrantTypes
...
+
     /// <summary>
     /// ReUse: the refresh token handle will stay the same when refreshing tokens
     /// OneTime: the refresh token handle will be updated when refreshing tokens
     /// </summary>
     public TokenUsage RefreshTokenUsage { get; set; } = TokenUsage.OneTimeOnly;
 
     /// <summary>
-    /// Gets or sets a value indicating whether the access token (and its claims) should be updated on a refresh token request.
+    /// Specifics whether the access token (and its claims) should be updated on a refresh token request.
```
brockallen, line 240: typo?

brockallen:
Hey Joe -- mark this as ready again once you want one final review. Thx.

src/IdentityServer/Extensions/ValidatedAuthorizeRequestExtensions.cs
```diff
@@ -165,8 +165,9 @@ private static NameValueCollection ToOptimizedRawValues(this ValidatedAuthorizeR
             {
                 // https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests 
                 // requires client id and response type to always be in URL
+                // REVIEW - Does this apply to pushed requests?
```
josephdecock, line 168: OIDC requires that when request objects are used, the client pass the client id and response type, in order to make the request a valid oauth request. With PAR, I don't think we need to worry about that, since PAR is an OAuth specification, and the requests we're making are valid PAR requests. In fact, an example in RFC 9126 pushes a request object containing a response_type claim without duplicating the response type outside the request object. 

Since OIDC is asking for this duplication in other circumstances in  order to be a valid oauth request, and the relevant oauth specification doesn't require it in this circumstance, I'm leaving it out.

josephdecock:
@brockallen, I've created new issues to track the follow up work on otel (#1445), and to consider cleaning up the `ToOptimized...` methods (possibly we can deprecate them along with the `IAuthorizationParametersMessageStore`?). 

Also, I'm convinced now that we don't need to duplicate the response_type parameter outside the request parameter when we use PAR with JAR. I don't think the special case in `ToOptimizedRawValues` applies here - see my [comment](https://github.com/DuendeSoftware/IdentityServer/pull/1424#discussion_r1367455980).

brockallen (approved):

leastprivilege:
grats! @josephdecock

Issue #1351, opened 2023-06-27
Consider EF ExecuteUpdate and SetProperty for our various cleanup processes


brockallen:
Seems there's a feature in EF7 where we don't have to load all the records to then update them. This might make some of our update and cleanup workflows work better.

josephdecock:
Docs for reference - https://learn.microsoft.com/en-us/ef/core/saving/execute-insert-update-delete

bbartels:
I created this PR to use ExecuteDelete during frequently used persistedgrant operations https://github.com/DuendeSoftware/IdentityServer/pull/1427

josephdecock:
In addition, we would also like to consider fuzzing the startup time of the cleanup job to mitigate the "thundering herd" issue, where all scaled instances start their job at once.

PR #1427, opened 2023-09-28
Reduces PersistedGrantStore DB roundtrips

bbartels:

We are running a HA setup of IdentityServer across multiple continents and experience significant latency during logins in sites that are geographically distant to the DB. We analysed the queries being sent to the DB during logins and found the following queries affecting PersistedGrants. 

```SQL
SELECT p."Id", p."ClientId", p."ConsumedTime", p."CreationTime", p."Data", p."Description", p."Expiration", p."Key", p."SessionId", p."SubjectId", p."Type"
FROM "PersistedGrants" AS p
WHERE p."Key" = @__token_Key_0

INSERT INTO "PersistedGrants" ("ClientId", "ConsumedTime", "CreationTime", "Data", "Description", "Expiration", "Key", "SessionId", "SubjectId", "Type")
VALUES (@p0, @p1, @p2, @p3, @p4, @p5, @p6, @p7, @p8, @p9)
RETURNING "Id";

SELECT p."Id", p."ClientId", p."ConsumedTime", p."CreationTime", p."Data", p."Description", p."Expiration", p."Key", p."SessionId", p."SubjectId", p."Type"
FROM "PersistedGrants" AS p
WHERE p."Key" = @__key_0

SELECT p."Id", p."ClientId", p."ConsumedTime", p."CreationTime", p."Data", p."Description", p."Expiration", p."Key", p."SessionId", p."SubjectId", p."Type"
FROM "PersistedGrants" AS p
WHERE p."Key" = @__key_0

DELETE FROM "PersistedGrants"
WHERE "Id" = @p0;
```
After:
```SQL
INSERT INTO "PersistedGrants" ("ClientId", "ConsumedTime", "CreationTime", "Data", "Description", "Expiration", "Key", "SessionId", "SubjectId", "Type")
VALUES (@p0, @p1, @p2, @p3, @p4, @p5, @p6, @p7, @p8, @p9)
RETURNING "Id";

SELECT p."Id", p."ClientId", p."ConsumedTime", p."CreationTime", p."Data", p."Description", p."Expiration", p."Key", p."SessionId", p."SubjectId", p."Type"
FROM "PersistedGrants" AS p
WHERE p."Key" = @__key_0

DELETE FROM "PersistedGrants"
WHERE "Id" = @p0;
```

This PR removes the need for the first SELECT (to verify if a persisted grant already exists) in the optimistic case. It does make the situation of the persisted grant already existing worse as it now needs 3 roundtrips instead of 2. But this should generally be rare as they are usually deleted as part of cleanup. This could potentially be improve further when EF Core gets around to adding an UpsertAsync API.

This PR also addresses the need for a SELECT in order to do a DELETE and replaces Remove/SaveChanges with ExecuteDelete

### DIFF
--- a/src/EntityFramework.Storage/Stores/PersistedGrantStore.cs
+++ b/src/EntityFramework.Storage/Stores/PersistedGrantStore.cs
@@ -120,24 +120,10 @@ public class PersistedGrantStore : Duende.IdentityServer.Stores.IPersistedGrantS
     public virtual async Task RemoveAsync(string key)
     {
         using var activity = Tracing.StoreActivitySource.StartActivity("PersistedGrantStore.Remove");
-        
-        var persistedGrant = (await Context.PersistedGrants.Where(x => x.Key == key)
-                .ToArrayAsync(CancellationTokenProvider.CancellationToken))
-            .SingleOrDefault(x => x.Key == key);
-        if (persistedGrant!= null)
+
+        if (await Context.PersistedGrants.ExecuteDeleteAsync(x => x.Key == key) > 0)
         {
-            Logger.LogDebug("removing {persistedGrantKey} persisted grant from database", key);
-
-            Context.PersistedGrants.Remove(persistedGrant);
-
-            try
-            {
-                await Context.SaveChangesAsync(CancellationTokenProvider.CancellationToken);
-            }
-            catch(DbUpdateConcurrencyException ex)
-            {
-                Logger.LogInformation("exception removing {persistedGrantKey} persisted grant from database: {error}", key, ex.Message);
-            }
+            Logger.LogDebug("removed {persistedGrantKey} persisted grant from database", key);
         }
         else
         {
--- a/src/EntityFramework.Storage/Stores/PersistedGrantStore.cs
+++ b/src/EntityFramework.Storage/Stores/PersistedGrantStore.cs
@@ -64,16 +64,16 @@ public class PersistedGrantStore : Duende.IdentityServer.Stores.IPersistedGrantS
             await Context.SaveChangesAsync(CancellationTokenProvider.CancellationToken);
             Logger.LogDebug("{persistedGrantKey} not found in database", token.Key);
         }
+        catch (DbUpdateConcurrencyException ex)
+        {
+            Logger.LogWarning("exception inserting or updating {persistedGrantKey} persisted grant in database: {error}", token.Key, ex.Message);
+        }
         catch (DbUpdateException)
         {
             Logger.LogDebug("{persistedGrantKey} found in database", token.Key);
             await Context.PersistedGrants.Where(x => x.Key == key).ExecuteDeleteAsync(CancellationTokenProvider.CancellationToken);
             Logger.LogDebug("removed {persistedGrantKey} persisted grant from database", key);
             await Context.SaveChangesAsync(CancellationTokenProvider.CancellationToken);
-        }
-        catch (DbUpdateConcurrencyException ex)
-        {
-            Logger.LogWarning("exception inserting or updating {persistedGrantKey} persisted grant in database: {error}", token.Key, ex.Message);
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/EntityFramework.Storage/Stores/PersistedGrantStore.cs
```diff
@@ -55,31 +55,25 @@ public PersistedGrantStore(IPersistedGrantDbContext context, ILogger<PersistedGr
...
         {
-            Logger.LogWarning("exception updating {persistedGrantKey} persisted grant in database: {error}", token.Key, ex.Message);
+            Logger.LogWarning("exception inserting or updating {persistedGrantKey} persisted grant in database: {error}", token.Key, ex.Message);
+        }
+        catch (DbUpdateException)
+        {
+            Logger.LogDebug("{persistedGrantKey} found in database", token.Key);
+            await Context.PersistedGrants.Where(x => x.Key == key).ExecuteDeleteAsync(CancellationTokenProvider.CancellationToken);
+            Logger.LogDebug("removed {persistedGrantKey} persisted grant from database", key);
+            await Context.SaveChangesAsync(CancellationTokenProvider.CancellationToken);
```
bbartels, line 76: This _could_ be done with ExecuteUpdate, but it does not seem to allow for updating the entire entity, but rather each property. Requiring any additional properties in the future to be included in the future -- so not as maintainable.

bbartels:
It does seem like this breaks the tests if running with the InMemory provider due to it being unable to translate the ExecuteDelete query. Switching over to Sqlite for tests resolves this issue.

bbartels:
I have also created https://github.com/bbartels/IdentityServer/pull/1, this PR comes with a few more side-effects, but would eliminate another database roundtrip. Would like to hear someone's thoughts on it.

bbartels:
Thinking about it, this may not be a safe change to make sadly. At least not in the current form. Anyone who currently supplies a `DbContext` backed by an `InMemory` provider to the `PersistedGrantStore` will have any tests broken that target the `TokenRequestValidator`. I feel like this almost requires a `ActualDbPersistedGrantStore` (name TBD) override (that could also do away with the abstraction over the dbcontext). Then users can opt into supplying `ActualDbPersistedGrantStore` instead of `PersistedGrantStore` if they want better database access performance.

Any thoughts?

josephdecock:
I actually think that we can allow the InMemory provider limitations. It does make this change a breaking change, but not a very severe one. Using the InMemory provider is [strongly discouraged in the documentation](https://learn.microsoft.com/en-us/ef/core/testing/choosing-a-testing-strategy#in-memory-as-a-database-fake), so I think it is reasonable for us to say, "We use features of EF that can't be tested with the InMemory provider".

josephdecock:
@AndersAbel - I think this is directly applicable to work that you're doing on #1351. Are you making similar changes/changes in the same place?

josephdecock:
@bbartels - after more consideration, we've decided to not merge these changes, but we do appreciate you opening a PR!

The reason we aren't going to merge this is that this code path is used by all persisted grants, and different persisted grants have very different access patterns. Depending on usage of IdentityServer overall, different implementations would see performance costs and benefits from your approach.

Our advice generally is that if you need to optimize data access, implementing the store interfaces that you want to optimize is the best thing to do. That gives you the ability to leverage everything you know about your access patterns and which database engine you're using. For example, you could optimize even further by using the "OUTPUT DELETED" feature in SQL Server, getting down to a single query per page of data. Optimizations like that aren't possible for us to implement, because we can't assume a particular database provider, but since you're in control of the provider, you absolutely can do that. You can also apply optimizations in a targeted way, based on your metrics. We have separate interfaces for each persisted grant type (IRefreshTokenStore, IAuthorizationCodeStore, IReferenceTokenStore, etc), to help you target your optimizations to the hottest paths.

Thanks again for your PR! Even though this isn't getting merged, your comments about testing with SqlLite sparked an interesting conversation and we are going to start doing that.

Issue #1433, opened 2023-10-05
Consider making ServerSideSessionStore.GetAndRemoveExpiredSessionsAsync virtual


AndersAbel:
Allows for customization using a derived class.

brockallen:
Is this from a customer request (that may be blocked), or did this just come up from your own casual observations?

AndersAbel:
It's a customer request.

brockallen:
Yes, this got released in a patch, then merged into main on Nov 1, 2023.

PR #1434, opened 2023-10-06
Make ServerSideSessionStore.GetAndRemoveExpiredSessionsAsync virtual

AndersAbel:
#1433


josephdecock:
Since you want to merge this into the 6.3 branch, our convention is to  not do a separate PR into main. Instead we will merge the 6.3 branch into main periodically. This makes it more obvious in the release notes when a fix occurred.

PR #1435, opened 2023-10-06
Make ServerSideSessionStore.GetAndRemoveExpiredSessionsAsync virtual

AndersAbel:
#1433

### DIFF
--- a/src/EntityFramework.Storage/Stores/ServerSideSessionStore.cs
+++ b/src/EntityFramework.Storage/Stores/ServerSideSessionStore.cs
@@ -262,7 +262,7 @@ public class ServerSideSessionStore : IServerSideSessionStore
 
 
     /// <inheritdoc/>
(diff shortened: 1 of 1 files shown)


brockallen (approved):

brockallen:
@AndersAbel do you know the urgency of releasing this (IOW, was a customer waiting)?

AndersAbel:
I don't think it's that urgent, they are still early in the implementation phase. I'll check with them.

josephdecock:
Did we ever decide if we were going to do a 6.3.6 release for this? It is in both 6.3.x and main, so it will go out in our 7.0 preview 2 this week.

AndersAbel:
I don't think there's any need to make a 6.3.6 just for this. I think it can wait for the next 6.3.x patch release.

brockallen:
IIRC we did release this into the 6.3.6 patch already.

josephdecock:
Yeah, I was confused because because it wasn't in our github releases, but that's fixed now.

PR #1453, opened 2023-10-30
Expose suppressed and original prompt modes

### LINKED ISSUE #1449: Expose original and suppressed prompt modes on validated authorize requests
6.2.1

7.0

**Describe the request**

We have an implementation of `ICustomAuthorizeRequestValidator` which contains some logic depending on PromptModes. 

Currently we use a workaround similar to the one introduced with https://github.com/DuendeSoftware/IdentityServer/pull/65. We are looking to remove our workaround and rely on the fix in IdentityServer. This works fine when accessing the promptModes from the returned result of `IIdentityServerInteractionService.GetAuthorizationContextAsync`, but the same is not true for the context passed into `ICustomAuthorizeRequestValidator`, as seen [here](https://github.com/DuendeSoftware/IdentityServer/blob/df2b00bb905905be27a25a3c00397c5020165ef6/src/IdentityServer/Validation/Default/AuthorizeRequestValidator.cs#L120C1-L132C55). `ValidateOptionalParametersAsync` mutates the PromptModes list just before passing it to the custom validator.  

The `ValidatedAuthorizeRequest` does have an `OriginalPromptModes`, but it is internal. Is it possible to make the promptModes available somehow? 

**To Reproduce**

Same as in https://github.com/IdentityServer/IdentityServer4/issues/5002, except instead of trying to access promptModes on the result of `IIdentityServerInteractionService.GetAuthorizationContextAsync`, try and access the prompmodes from within a `ICustomAuthorizeRequestValidator`.

josephdecock:
Closes #1449


src/IdentityServer/Validation/Models/ValidatedAuthorizeRequest.cs
```diff
@@ -138,26 +138,66 @@ public class ValidatedAuthorizeRequest : ValidatedRequest
...
+    /// <list type="bullet">
+    /// <item><seealso cref="PromptModes"/></item>
+    /// <item><seealso cref="OriginalPromptModes"/></item>
+    /// </list>
+    /// </para>
     /// <value>
     /// The collection of suppressed prompt modes.
     /// </value>
-    internal IEnumerable<string> SuppressedPromptModes { get; set; } = Enumerable.Empty<string>();
+    public IEnumerable<string> SuppressedPromptModes { get; set; } = Enumerable.Empty<string>();
```
brockallen, line 200: Maybe we want a better name for these, like "ProcessedPromptModes"? Thoughts? Suggestions?

src/IdentityServer/Constants.cs
```diff
@@ -115,7 +115,7 @@ public static class SigningAlgorithms
         //OidcConstants.PromptModes.Create, 
     };
 
-    public const string SuppressedPrompt = "suppressed_" + OidcConstants.AuthorizeRequest.Prompt;
+    public const string ProcessedPrompt = "processed_" + OidcConstants.AuthorizeRequest.Prompt;
```
josephdecock, line 118: Renaming here makes the names consistent, but is a (very minor) breaking change.

brockallen: This would only break someone in-flight, right?

josephdecock: Yeah, if you're using the constant or looking at the parameters raw and using the old name, this means you need to make an update.

src/IdentityServer/Validation/Models/ValidatedAuthorizeRequest.cs
brockallen: Ok, let's revert the constant and keep the new "Processed" name on the new prop, please.

src/IdentityServer/Constants.cs
brockallen: Leave the old constant and deprecate it w/ obsolete, please.

josephdecock:
@brockallen this is ready for review again.

src/IdentityServer/Constants.cs
```diff
@@ -115,7 +115,21 @@ public static class SigningAlgorithms
...
     };
 
-    public const string ProcessedPrompt = "processed_" + OidcConstants.AuthorizeRequest.Prompt;
+    /// <summary>
+    /// The name of the parameter passed to the authorize callback to indicate
+    /// prompt modes that have already been used. This constant is deprecated in
+    /// favor of <see cref="ProcessedPrompt"/>.
+    /// </summary>
+    [Obsolete("Use the ProcessedPrompt constant instead.")]
+    public const string SuppressedPrompt = "suppressed" + OidcConstants.AuthorizeRequest.Prompt;
```
brockallen, line 124: we lost the "_", no?

josephdecock: Thanks for catching this - fixed!

brockallen (approved):

brockallen:
@josephdecock darn, now some conflicts

PR #1454, opened 2023-11-01
Provide abstraction for IdentityServerTools

matt-goldman:
`IdentityServerTools` is useful and I expect used in a lot of places. However as it's a concrete class it is difficult to test services that depend on it. This PR creates an abstraction `IIdentityServerTools` which can be injected instead, and can therefore be mocked easily for tests. This is important for testing services that depend upon `IdentityServerTools` but don't explicitly need it for testing purposes.

**Note:** This would be a breaking change. Personally I think that's ok, but if you wanted to maintain backward compatibility, we could keep both registrations:

```csharp
builder.Services.AddTransient<IIdentityServerTools, IdentityServerTools>();
builder.Services.AddTransient<IdentityServerTools>();
```

### DIFF
--- a/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/Core.cs
+++ b/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/Core.cs
@@ -190,7 +190,13 @@ public static class IdentityServerBuilderExtensionsCore
         builder.Services.AddTransient<IJwtRequestValidator, JwtRequestValidator>();
 
         builder.Services.AddTransient<ReturnUrlParser>();
+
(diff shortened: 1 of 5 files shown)


brockallen:
@AndersAbel check in our docs for updates based on this: https://docs.duendesoftware.com/identityserver/v6/tokens/internal/

brockallen:
Thanks for this @matt-goldman -- we'll work on it!

AndersAbel:
If someone has customized the service/class and replaced it with their own in DI this will revert back to using the default version of the service. I added Obsolete because I think that helps flagging this for anyone that has usage/customization.

AndersAbel (approved):

PR #1457, opened 2023-11-01
Cleanup deprecated code and TODOs

josephdecock:
- Remove code marked as unused/to be removed in 7.0
- Fix typo in name
- Clean up some very old todos that don't need to be done anymore


src/Storage/Constants.cs
```diff
@@ -6,6 +6,6 @@ namespace Duende.IdentityServer;
 
 internal static class Constants
 {
-    public const string IdentityServerName               = "IdentityServer4";
+    public const string IdentityServerName               = "Duende.IdentityServer";
```
brockallen, line 9: where is this used? will this change cause any breaking changes?

josephdecock: I don't think this is likely to break anyone. It is used
- As the default value for the CorsPolicyName option and
- As the value for IdentityServerAuthenticationType

I guess theoretically if someone has created a cors policy and named it "Duende.IdentityServer", this conflicts with that. That seems pretty unlikely to me.

And theoretically, a customer _could_ have a customization that cares about the authentication type and refers to it without using this constant. Again, seems pretty unlikely.

OTOH, this is a low-value change.

brockallen (approved):

PR #1458, opened 2023-11-02
Allow arbitrary pushed redirect_uris

josephdecock:
Spec allows an arbitrary redirect uri to be pushed: https://datatracker.ietf.org/doc/html/rfc9126#name-management-of-client-redire

This is implemented as a new redirect uri validator that is passed the necessary context to determine if redirection is needed.

Question: should this behavior always be enabled, or be opt in?


src/IdentityServer/Validation/Models/ValidatedAuthorizeRequest.cs
```diff
@@ -221,9 +221,11 @@ public class ValidatedAuthorizeRequest : ValidatedRequest
     public string? PushedAuthorizationReferenceValue { get; set; }
 
     /// <summary>
-    /// Gets a value indicator if the authorize request's parameters where pushed using PAR. 
+    /// Gets or sets a value indicating the context in which authorization
+    /// validation is occurring (the PAR endpoint or the authorize endpoint with
+    /// or without pushed parameters).
     /// </summary>
-    public bool IsPushedAuthorizationRequest => PushedAuthorizationReferenceValue is not null;
+    public AuthorizeRequestType AuthorizeRequestType { get; set; }
```
brockallen, line 228: Maybe make it overly explicit that it's set to `Authorize` by default?

brockallen: Meh, I now see this is used in many places... we can just rely upon the default enum values, I suppose.

brockallen:
Let's schedule a call to pair review this, please.

josephdecock:
This is ready for another review.

src/IdentityServer/Validation/Default/DefaultClientConfigurationValidator.cs
```diff
@@ -131,7 +131,11 @@ protected virtual Task ValidateRedirectUriAsync(ClientConfigurationValidationCon
                 context.Client.AllowedGrantTypes.Contains(GrantType.Hybrid) ||
                 context.Client.AllowedGrantTypes.Contains(GrantType.Implicit))
             {
-                if (context.Client.RedirectUris?.Any() == false)
+                // Clients must have redirect uris, unless they use PAR since it
+                // is valid to use a pushed redirect uri that is not
+                // pre-arranged.
+                if (context.Client.RedirectUris?.Any() == false && 
+                    !context.Client.RequirePushedAuthorization)
```
brockallen, line 138: Confidential, not require par, right?

src/IdentityServer/Validation/Default/StrictRedirectUriValidator.cs
```diff
@@ -17,6 +18,17 @@ namespace Duende.IdentityServer.Validation;
 /// </summary>
 public class StrictRedirectUriValidator : IRedirectUriValidator
 {
+    private readonly IdentityServerOptions _options;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="StrictRedirectUriValidator" />.
+    /// </summary>
+    /// <param name="options"></param>
+    public StrictRedirectUriValidator(IdentityServerOptions options)
```
brockallen, line 27: Maybe optional for backwards compat, and if null then we assume no PAR?

src/IdentityServer/Validation/Default/DefaultClientConfigurationValidator.cs
josephdecock: Right, forgot about this one!

src/IdentityServer/Validation/Default/StrictRedirectUriValidator.cs
josephdecock: I went with the optional parameters here (and similarly in the app auth ctor)

brockallen (approved):

brockallen:
👍

PR #1461, opened 2023-11-06
Relax handling of DCR license issues

josephdecock:
Instead of shutting down if the license is wrong, warn. This makes DCR behave the same as e.g., automatic key management.

### DIFF
--- a/src/Configuration/Licensing/ConfigurationLicenseValidator.cs
+++ b/src/Configuration/Licensing/ConfigurationLicenseValidator.cs
@@ -30,7 +30,7 @@ internal class ConfigurationLicenseValidator : LicenseValidator<ConfigurationLic
         
         if (!License.ConfigApiFeature)
         {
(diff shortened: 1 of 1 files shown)


josephdecock:
I noticed this while testing PAR licensing, and I don't think this is what we originally intended.

brockallen:
>  I don't think this is what we originally intended.

I thought we made a move to prevent use of higher tier features so people would not be surprised that their license did not support it. Some feature we left in the more relaxed mode, but look here:

https://github.com/DuendeSoftware/IdentityServer/blob/main/src/IdentityServer/Licensing/IdentityServerLicenseValidator.cs#L132
https://github.com/DuendeSoftware/IdentityServer/blob/main/src/IdentityServer/Licensing/IdentityServerLicenseValidator.cs#L149
https://github.com/DuendeSoftware/IdentityServer/blob/main/src/IdentityServer/Licensing/IdentityServerLicenseValidator.cs#L168

etc

in short -- if there is a license and they're using a whole feature that's not in the license then we need to fail.

PR #1462, opened 2023-11-07
Discourage IAuthorizationParametersMessageStore

### LINKED ISSUE #1444: Clean up ValidatedAuthorizeRequest extension methods (ToOptimized...)
`ToOptimizedQueryString` and `ToOptimizedFullDictionary` are only used when we have the `IAuthorizationParametersMessageStore`. Consider deprecating all of it, or at least the `ToOptimized...` methods.

josephdecock:
PAR is a more robust/standardized approach to get similar benefits

This PR obsoletes the extension method that registers the default `IAuthorizationParametersMessageStore` as well as the `ToOptimizedFullDictionary` method, which is only used when that store is used. There are more types that could in theory be obsoleted - notably the `IAuthorizationParametersMessageStore` itself, but that gets used in many places, and would add quite a bit of clutter.

This closes #1444

### DIFF
--- a/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/Additional.cs
+++ b/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/Additional.cs
@@ -492,6 +492,7 @@ public static class IdentityServerBuilderExtensionsAdditional
     /// <typeparam name="T"></typeparam>
     /// <param name="builder">The builder.</param>
     /// <returns></returns>
+    [Obsolete("This feature is deprecated. Consider using Pushed Authorization Requests instead.")]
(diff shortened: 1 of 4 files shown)


brockallen (approved):

wcabus:
While I'd love to use PAR everywhere instead of passing a bunch of query parameters to the authorize endpoint, it's not really feasible to expect that every client suddenly can switch to using PAR. Our own client applications can be updated over time, but unfortunately, we can't force or control external applications that also use our Duende instance.

And because we had some issues with URL's becoming too large, we are using the `DistributedCacheAuthorizationParametersMessageStore` backed by a distributed SQL database cache.

PR #1467, opened 2023-11-10
Otel metrics for PAR

### LINKED ISSUE #1445: Add Otel metrics to PAR implementation

AndersAbel:
Fixes #1445

### DIFF
--- a/src/IdentityServer/Endpoints/PushedAuthorizationEndpoint.cs
+++ b/src/IdentityServer/Endpoints/PushedAuthorizationEndpoint.cs
@@ -84,12 +84,17 @@ internal class PushedAuthorizationEndpoint : IEndpointHandler
 
         var response = await _responseGenerator.CreateResponseAsync(parValidationResult.ValidatedRequest);
 
-        return response switch
+        switch(response)
         {
-            PushedAuthorizationSuccess success => new PushedAuthorizationResult(success),
-            PushedAuthorizationFailure fail => new PushedAuthorizationErrorResult(fail),
-            _ => throw new Exception("Can't happen")
-        };
+            case PushedAuthorizationSuccess success:
+                Telemetry.Metrics.PushedAuthorizationRequest(parValidationResult.ValidatedRequest.Client.ClientId);
+                return new PushedAuthorizationResult(success);
+            case PushedAuthorizationFailure fail:
+                Telemetry.Metrics.PushedAuthorizationRequestFailure(parValidationResult.ValidatedRequest.ClientId, fail.Error);
+                return new PushedAuthorizationErrorResult(fail);
+            default:
+                throw new Exception("Can't happen");
+        }
     }
 
     private PushedAuthorizationErrorResult CreateErrorResult(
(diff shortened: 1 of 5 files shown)


src/IdentityServer/Endpoints/PushedAuthorizationEndpoint.cs
```diff
@@ -84,12 +84,17 @@ public async Task<IEndpointResult> ProcessAsync(HttpContext context)
...
-            _ => throw new Exception("Can't happen")
-        };
+            case PushedAuthorizationSuccess success:
+                Telemetry.Metrics.PushedAuthorizationRequest(parValidationResult.ValidatedRequest.Client.ClientId);
+                return new PushedAuthorizationResult(success);
+            case PushedAuthorizationFailure fail:
+                Telemetry.Metrics.PushedAuthorizationRequestFailure(parValidationResult.ValidatedRequest.ClientId, fail.Error);
+                return new PushedAuthorizationErrorResult(fail);
+            default:
+                throw new Exception("Can't happen");
```
brockallen, line 96: @josephdecock maybe while reviewing this, you can consider a better exception message here?

brockallen:
@josephdecock since this touches your recent additions, i think it's best for you to review it please. thanks!

src/IdentityServer/Validation/Models/ValidatedAuthorizeRequest.cs
```diff
@@ -259,6 +260,11 @@ public class ValidatedAuthorizeRequest : ValidatedRequest
     /// </summary>
     public string? PushedAuthorizationReferenceValue { get; set; }
 
+    /// <summary>
+    /// Is this a pushed authorization request?
+    /// </summary>
+    public bool IsPushedAuthorizationRequest { get => PushedAuthorizationReferenceValue.IsPresent(); }
```
josephdecock, line 266: There is already an AuthorizeRequestType on the validated request that uses an enum to distinguish 3 states:
1. PAR
2. Authorize with Pushed Parameters
3. Authorize without Pushed Parameters

I don't know if it makes sense to include that distinction in any of these metrics - what do you think?

Regardless, my preference would be to use that existing property instead of introducing another flag. Earlier on in development of PAR, I only had a flag instead of the enum and confused myself.

src/IdentityServer/Endpoints/PushedAuthorizationEndpoint.cs
josephdecock: Yeah, good idea - this is done

src/IdentityServer/Validation/Models/ValidatedAuthorizeRequest.cs
AndersAbel: That's why I didn't find the property you mentioned in https://github.com/DuendeSoftware/IdentityServer/issues/1445#issuecomment-1775318325. I got a bit confused... I'll change to use the enum.

josephdecock (approved):

PR #1484, opened 2023-12-11
Fix incorrect LogWarning call

AndersAbel:
The exception should be the first param. The existing code gives a build warning that the number of params doesn't match the formatting string.


brockallen (approved):

AndersAbel:
@brockallen Yes.

PR #1486, opened 2023-12-13
Fix logout token iss when issuer is missing

### LINKED ISSUE #1481: Back channel logout token might incorrectly contain `logout_token` as `iss`
If the request does not contain an issuer, the issuer name service should be invoked to get the issuer. This is done automatically in the two param version of IIdentityServerTools.IssueJwtAsync.

https://github.com/DuendeSoftware/IdentityServer/pull/1170 adds support for setting `typ` to `logout+jwt`. Unfortunately the code path with no issuer in the request now mistakenly uses the three param version of the constructor. This results in the token type `logout_token` being used as `iss`

josephdecock:
Fixes #1481

### DIFF
--- a/src/IdentityServer/IdentityServerTools.cs
+++ b/src/IdentityServer/IdentityServerTools.cs
@@ -19,7 +19,7 @@ namespace Duende.IdentityServer;
 /// </summary>
 public class IdentityServerTools
 {
-    internal readonly IServiceProvider ServiceProvider;
+    internal readonly IServiceProvider ServiceProvider; // TODO - consider removing this, as it is not used.
     internal readonly IIssuerNameService IssuerNameService;
     private readonly ITokenCreationService _tokenCreation;
     private readonly ISystemClock _clock;
--- a/src/IdentityServer/IdentityServerTools.cs
+++ b/src/IdentityServer/IdentityServerTools.cs
@@ -48,23 +48,22 @@ public class IdentityServerTools
     /// <exception cref="System.ArgumentNullException">claims</exception>
     public virtual async Task<string> IssueJwtAsync(int lifetime, IEnumerable<Claim> claims)
     {
-        var tokenType = OidcConstants.TokenTypes.AccessToken;
         var issuer = await IssuerNameService.GetCurrentAsync();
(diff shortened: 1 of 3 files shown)


src/IdentityServer/IdentityServerTools.cs
```diff
@@ -48,22 +48,23 @@ public IdentityServerTools(IServiceProvider serviceProvider, IIssuerNameService
...
     /// Issues a JWT.
     /// </summary>
     /// <param name="lifetime">The lifetime.</param>
-    /// <param name="issuer">The issuer.</param>
+    /// <param name="tokenType">The token type.</param>
     /// <param name="claims">The claims.</param>
     /// <returns></returns>
     /// <exception cref="System.ArgumentNullException">claims</exception>
-    public virtual Task<string> IssueJwtAsync(int lifetime, string issuer, IEnumerable<Claim> claims)
+    public virtual async Task<string> IssueJwtAsync(int lifetime, string tokenType, IEnumerable<Claim> claims)
```
josephdecock, line 64: Oh wait, this is a breaking change. Need to fix the caller instead I think.

josephdecock: To fix the caller, we have to add a dependency on the `IssuerNameService` from the `DefaultBackChannelLogoutService`, which technically is also a breaking change. Maybe we don't need this in a 6.3 patch release?

brockallen: If it's a bug and not working as expected, then I don't think it's a problem to make a breaking change.

josephdecock: Okay, then I'll keep this for the 6.3 branch, but still rework the fix slightly so that we're not changing semantics of the method in IdentityServerTools. This way the "breaking" change only affects the bug, and won't surprise people (the other change would probably have compiled without an error!)

brockallen (approved):

josephdecock:
Note that we now rely on issuer name service, which indirectly relies on http context, so it is possible that server side session expiry will fail, unless an issuer is explicitly configured.

PR #1489, opened 2023-12-14
Hex encode consent keys

### LINKED ISSUE #1325: Use hex encoding for consent keys
Beginning in v6.0, the hashes passed to the `IPersistedGrantStore` to use as `Key` values changed to be formatted as hex values. In earlier versions, the keys were base-64 encoded. This occasionally caused issues when the data store used case-insensitive collation, which prompted the change to hex encoding. To facilitate migration, IdentityServer adds a version suffix ("-1") to the parameter values to indicate that the newer hex encoding should be used.  However, consent records were not migrated to use hex encoding of their `Key` values. Since there's no protocol parameter that corresponds to consent records, there's no way to use the protocol parameters to determine which encoding to use. So currently, the consent `Key` values have continued to use base-64 encoding. 

However, there is a theoretical risk of the same database collation issues affecting consents. If two consent hashes differ only by casing in the base-64 encodings, then I believe the EF persisted grant store would fail to store the second grant:

```cs
        var existing = (await Context.PersistedGrants.Where(x => x.Key == token.Key)
                .ToArrayAsync(CancellationTokenProvider.CancellationToken))
            .SingleOrDefault(x => x.Key == token.Key);

        // This will be true, because of the second in-memory SingleOrDefault call
        if (existing == null)
        {
            Logger.LogDebug("{persistedGrantKey} not found in database", token.Key);

            var persistedGrant = token.ToEntity();

            // So, we will attempt to add the new grant to the table
            Context.PersistedGrants.Add(persistedGrant); 
        }
        else
        {
            Logger.LogDebug("{persistedGrantKey} found in database", token.Key);

            token.UpdateEntity(existing);
        }

        try
        {

            // ... however, this will fail if the database using a case-insensitive collation
            await Context.SaveChangesAsync(CancellationTokenProvider.CancellationToken); 
        }
        catch (DbUpdateConcurrencyException ex)
        {
            Logger.LogWarning("exception updating {persistedGrantKey} persisted grant in database: {error}", token.Key, ex.Message);
        }
```

We can prevent this by using hex encoding for Consent `Key` values. To facilitate migration, we'll need to first query using hex encoding, and then fall back to base-64 encoding if hex encoding fails to find a grant. If we find a grant with the fallback method, then we should update the grant to hex encoding.

josephdecock:
Resolves #1325

### DIFF
--- a/src/IdentityServer/Stores/Default/DefaultGrantStore.cs
+++ b/src/IdentityServer/Stores/Default/DefaultGrantStore.cs
@@ -72,7 +72,11 @@ public class DefaultGrantStore<T>
     }
 
     private const string KeySeparator = ":";
-    const string HexEncodingFormatSuffix = "-1";
+    
+    /// <summary>
+    /// The suffix added to keys to indicate that hex encoding should be used.
+    /// </summary>
+    protected const string HexEncodingFormatSuffix = "-1";
 
     /// <summary>
     /// Creates a handle.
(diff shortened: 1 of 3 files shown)


brockallen (approved):

josephdecock:
Migrating records is needed because we often re-request consent (if offline_access is requested).

PR #1490, opened 2023-12-15
Brock/merge 6.3.7

brockallen:


josephdecock (approved):

josephdecock (changes requested):
Actually, I want to revisit the use of the service provider...

josephdecock:
> Actually, I want to revisit the use of the service provider...

On second thought, the service provider should be done on its own. This is good merge.

josephdecock (approved):

PR #1491, opened 2023-12-15
Move tools extension method into tools itself

josephdecock:
This simplifies the new interface considerably. And note that the new interface was introduced for v7, so this isn't a breaking change, apart from adding a new dependency to the ctor.


test/IdentityServer.UnitTests/Services/Default/DefaultBackChannelLogoutServiceTests.cs
```diff
@@ -57,10 +57,10 @@ public async Task CreateTokenAsync_Should_Set_Issuer_Correctly()
         var issuerNameService = new TestIssuerNameService(expected);
 #pragma warning disable CS0618 // Type or member is obsolete
```
brockallen, line 58: we can remove these pragmas then too

josephdecock: Yeah, in the other PR I think.

josephdecock:
@matt-goldman, this seems related to your pr: #1454.  Does refactoring in this way work for you?

matt-goldman:
Does it make sense to reflect these changes in the `IIdentityServerTools` interface, given that they're not in the extensions interface now?

josephdecock:
> Does it make sense to reflect these changes in the `IIdentityServerTools` interface, given that they're not in the extensions interface now?

Yes, absolutely, and Thanks!

PR #1492, opened 2023-12-15
Remove [Obsolete] from IdentityServerTools

josephdecock:
`[Obsolete]` was added to the `IdentityServerTools` class when we introduced the `IIdentityServerTools` interface. This was added to try to alert users that they may need to change their DI registrations.

Users can customize `IdentityServerTools` today by deriving from the class, and then registering the custom version like this:
```cs
builder.Services.AddTransient<IdentityServerTools, CustomIdentityServerTools>
```

Pre-v7, this works because our code depends on an `IdentityServerTools`, which resolves to `CustomIdentityServerTools`. In v7, this fails because our code depends on an `IIdentityerServerTools`, which resolves to `IdentityServerTools`.

Users need to change to 
```cs
builder.Services.AddTransient<IIdentityerServerTools, CustomIdentityServerTools>
```

Marking `IdentityServerTools` as `[Obsolete]` provides a compiler warning to draw attention to this issue.

However, the downside to this is that now `CustomIdentityServerTools` will complain that its base class is obsolete, even though it is still valid to derive from `IdentityServerTools` and override its virtual methods. 

Either way can cause confusion, but I think its better to treat this as a documentation/release notes problem, rather than trying to use the bend the semantics of this compiler warning this far.

### DIFF
--- a/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/Core.cs
+++ b/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/Core.cs
@@ -188,16 +188,8 @@ public static class IdentityServerBuilderExtensionsCore
         builder.Services.AddTransient<ExtensionGrantValidator>();
         builder.Services.AddTransient<BearerTokenUsageValidator>();
         builder.Services.AddTransient<IJwtRequestValidator, JwtRequestValidator>();
-
         builder.Services.AddTransient<ReturnUrlParser>();
-
-#pragma warning disable CS0618 // Type or member is obsolete
         builder.Services.AddTransient<IIdentityServerTools, IdentityServerTools>();
-        // We've added the IIdentityServerTools interface to allow mocking, but keep the old
-        // direct class registration around if anyone has a dependency on it.
-        builder.Services.AddTransient<IdentityServerTools>();
-#pragma warning restore CS0618 // Type or member is obsolete
-
         builder.Services.AddTransient<IReturnUrlParser, OidcReturnUrlParser>();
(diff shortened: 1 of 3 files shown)


brockallen (approved):

brockallen:
Let's ping the customer who made the original request for the IIdentityServerTools change to have them have a look?

josephdecock:
> Let's ping the customer who made the original request for the IIdentityServerTools change to have them have a look?

@matt-goldman, thanks for your original PR that this is a follow-up to. Any feedback on this?

josephdecock:
Also @AndersAbel, I think you had the idea of using the `[Obsolete]` attribute. It's a cool trick to get users to fix their DI registrations, but I'm worried it is a little _too_ tricky. Thoughts?

AndersAbel:
> Also @AndersAbel, I think you had the idea of using the `[Obsolete]` attribute. It's a cool trick to get users to fix their DI registrations, but I'm worried it is a little _too_ tricky. Thoughts?

I think the reason for Obsolete was to give anyone referencing the `IdentityServerTools` a heads up that things changed. This was relevant because we kept the `IdentityServerTools` class registered to not break anyone using it. But if someone registered a custom version of the tools IdentityServer would now silently not accept that.

But in the end, using `[Obsolete]` is probably a bit _too_ tricky. It's a major version, we can do breaking changes so let's remove `[Obsolete]`. But I think we should then also remove `AddTransient<IdentityServerTools>()`.

matt-goldman:
I agree that breaking changes are fine with major versions, as long as they're suitably documented.

I'm not 100% sure about inheriting `IdentityServerTools` to customise it...but that's not my project anyway :)

PR #1493, opened 2023-12-15
implement correct overload for IsRedirectUriValidAsync in StrictRedirectUriValidatorAppAuth

brockallen:


josephdecock (approved):

josephdecock:
This fixes a regression that was introduced in the PAR work in 7.0.0 preview 2, where the AppAuth redirect uri validator wouldn't actually be invoked.

PR #1497, opened 2023-12-19
CIBA Customization Extensibility Points

josephdecock:
This adds extensibility points to CIBA to allow for custom request and response parameters.

- Adds a new custom backchannel authentication request validator hook, similar to `AddCustomAuthorizeRequestValidator` and `AddCustomTokenRequestValidator`
- Adds a context dictionary to the ciba validation result, which then feeds into similar context dictionaries in the response to the client, notification to the user, and record in the database

### DIFF
--- a/src/IdentityServer/Models/BackchannelUserLoginRequest.cs
+++ b/src/IdentityServer/Models/BackchannelUserLoginRequest.cs
@@ -62,8 +62,8 @@ public class BackchannelUserLoginRequest
     public ResourceValidationResult ValidatedResources { get; set; } = default!;
 
     /// <summary> 
-    /// Gets or sets a dictionary of custom properties that can pass extra state
-    /// to the notification process.
+    /// Gets or sets a dictionary of custom properties that can pass additional
+    /// state to the notification process.
     /// </summary>
-    public Dictionary<string, object> Context { get; set; } = new();
+    public Dictionary<string, object> Properties { get; set; } = new();
 }
--- a/src/IdentityServer/Models/BackchannelUserLoginRequest.cs
+++ b/src/IdentityServer/Models/BackchannelUserLoginRequest.cs
@@ -61,8 +61,9 @@ public class BackchannelUserLoginRequest
     /// </summary>
     public ResourceValidationResult ValidatedResources { get; set; } = default!;
 
-    /// <summary>
-    /// The response that will be sent to the client
+    /// <summary> 
+    /// Gets or sets a dictionary of custom properties that can pass extra state
+    /// to the notification process.
     /// </summary>
-    public BackchannelAuthenticationResponse Response { get; set; } = default!;
+    public Dictionary<string, object> Context { get; set; } = new();
 }
--- a/src/IdentityServer/Models/BackchannelUserLoginRequest.cs
+++ b/src/IdentityServer/Models/BackchannelUserLoginRequest.cs
@@ -7,6 +7,7 @@
 using Duende.IdentityServer.Validation;
 using System.Collections.Generic;
 using System.Security.Claims;
+using Duende.IdentityServer.ResponseHandling;
 
 namespace Duende.IdentityServer.Models;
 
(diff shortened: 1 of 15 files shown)


josephdecock:
This PR began as #1361, but that PR didn't get merged for a long time, so I rebased it and created a new PR here. @gislikonrad - thanks for raising this issue and submitting the PR back in the summer! Do you think this new approach will meet your needs?

brockallen:
Two thoughts:

1) I like the concept that it's a "context", but it's also "custom". So maybe a better name to indicate that it's custom? The token EP has "Custom", but maybe "CustomContext"?

2) We don't have any tests for this new feature. We should (both for the new custom validator extensibility point, but also emitting custom results that flatten into the response payload).

src/IdentityServer/Validation/Models/ValidatedBackchannelAuthenticationRequest.cs
```diff
@@ -83,4 +83,10 @@ public class ValidatedBackchannelAuthenticationRequest : ValidatedRequest
     /// Gets or sets the request object (either passed by value or retrieved by reference)
     /// </summary>
     public string? RequestObject { get; set; }
+
+    /// <summary>
+    /// Gets or sets a dictionary of custom properites that can store
+    /// additional state during the back channel authentication process.
+    /// </summary>
+    public Dictionary<string, object> Context { get; set; } = new();
```
gislikonrad, line 91: I feel `Context` isn't a good name for this. `Properties` would maybe be better.

src/Storage/Models/BackChannelAuthenticationRequest.cs
```diff
@@ -87,7 +87,12 @@ public class BackChannelAuthenticationRequest
...
     /// <summary>
-    /// Gets the description the user assigned to the client being authorized.
+    /// Gets or sets the description the user assigned to the client being authorized.
     /// </summary>
     public string? Description { get; set; }
+
+    /// <summary>
+    /// Gets or sets a dictionary of custom properties associated with this instance.
+    /// </summary>
+    public Dictionary<string, object> Context { get; set; } = new();
```
gislikonrad, line 97: `Context` isn't a good name for this. `Properties` would maybe be better. This is because it would be an analog to the `Properties` dictionary in `AuthorizationCode`.

Also, I would be careful using a dictionary with `object` values in a storage model. Maybe using `string` values would be enough for this and all other `Properties` dictionaries.

src/IdentityServer/Models/BackchannelUserLoginRequest.cs
```diff
@@ -59,4 +60,10 @@ public class BackchannelUserLoginRequest
     /// Gets or sets the validated resources.
     /// </summary>
     public ResourceValidationResult ValidatedResources { get; set; } = default!;
+
+    /// <summary> 
+    /// Gets or sets a dictionary of custom properties that can pass extra state
+    /// to the notification process.
+    /// </summary>
+    public Dictionary<string, object> Context { get; set; } = new();
```
gislikonrad, line 68: I would change this to `Properties`. I feel **Context** should be reserved for a property containing a context object.

gislikonrad (commented):
This would work for us at our organization. I just had some comments on the naming of the new properties in the backchannel authentication models and contexts.

src/Storage/Models/BackChannelAuthenticationRequest.cs
josephdecock: We use a dictionary<string, object> for the [JsonExtensionData], because it is required there. In the context of the storage model, we could use strings instead, but then we'd have to convert to and from strings as we pass the properties around, and if folks want to put complex objects into the properties, that adds some complexity. 

In the end, the dictionaries get serialized to json when they're stored. If you try to pass something that doesn't serialize, I would expect an error from the serializer, so I think the object is fine.

gislikonrad: Fair enough

brockallen (approved):

PR #1515, opened 2024-02-06
Open Telemetry Fixes

AndersAbel:
Rework the naming structure of the counters as the original names didn't follow the open telemetry conventions.

### DIFF
--- a/src/Telemetry/Telemetry.cs
+++ b/src/Telemetry/Telemetry.cs
@@ -272,7 +272,7 @@ public static class Telemetry
         {
             if (clientId != null)
             {
-                SuccessCounter.Add(1, tag: new("client", clientId));
+                SuccessCounter.Add(1, tag: new(Tags.Client, clientId));
             }
             else
             {
@@ -294,11 +294,11 @@ public static class Telemetry
         {
             if (clientId != null)
             {
-                FailureCounter.Add(1, new("client", clientId), new("error", error));
+                FailureCounter.Add(1, new(Tags.Client, clientId), new(Tags.Error, error));
             }
             else
             {
-                FailureCounter.Add(1, tag: new("error", error));
+                FailureCounter.Add(1, tag: new(Tags.Error, error));
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/Telemetry/Telemetry.cs
```diff
@@ -21,223 +22,81 @@ public static class Telemetry
...
-
-            /// <summary>
-            /// unhandled_exception
-            /// </summary>
-            public const string UnHandledException = "unhandled_exception";
+            public const string Operation = "tokenservice.operation";
+            public const string ActiveRequests = "tokenservice.active_requests";
+            public const string ApiSecretValidation = "tokenservice.api.secret_validation";
+            public const string BackchannelAuthentication = "tokenservice.backchannel_authentication";
+            public const string ClientValidation = "tokenservice.client.config_validation";
```
Tornhoof, line 46: Rename ClientValidation to ClientConfigValidation?

src/Telemetry/Telemetry.cs
```diff
@@ -21,223 +22,81 @@ public static class Telemetry
...
-            /// </summary>
-            public const string TokenIssuedFailure = "token_issued_failure";
-
-            /// <summary>
-            /// unhandled_exception
-            /// </summary>
-            public const string UnHandledException = "unhandled_exception";
+            public const string Operation = "tokenservice.operation";
+            public const string ActiveRequests = "tokenservice.active_requests";
+            public const string ApiSecretValidation = "tokenservice.api.secret_validation";
```
Tornhoof, line 44: I'm not sure, that I understand the logic between . and _ in some of the names:
Assuming . is similar to a namespace, then it should be both .
"tokenservice.device_authentication" has an _
"tokenservice.resourceowner.authentication" has a .

AndersAbel: @Tornhoof The `.` is a namespace separator, the `_` is a word separator. Some tools creates a drill-down menu based on the dots. `device_authentication` and `resourceowner.authentication` should probably be the same, thanks for noticing.

brockallen (approved):

PR #1521, opened 2024-02-15
Filter subject token from TokenRequest log

krosn:
Treat the Subject Token as sensitive because it could be an access token or an identity token that contains PII.

See https://github.com/DuendeSoftware/IdentityServer/issues/1522


brockallen:
Thanks @krosn. I'd want to consider if anyone cared/wanted it in the logs... of course, we can document this change. But we'd need to at least do it in a point release. So prolly in our 7.1.0 release.

krosn:
Sure, sounds good. There's no rush on my end.

josephdecock (approved):

PR #1549, opened 2024-05-06
Bump OpenTelemetry dependencies

AndersAbel:
Most of the packages are listed with a security vulernability that is only fixed in 1.8.1, so all them are bumped to that version. Except those packages that do not have a 1.8.1 released


brockallen (approved):

AndersAbel:
We need to get this into the releases/7.0.x branch too. I changed the target of #1550 and now it doesn't compile due to this.

Would we even need the fix into any of the 6.x release branches?

@brockallen @josephdecock

PR #1550, opened 2024-05-06
Anders/warn non distributed cache

AndersAbel:
Log (info level) if OIDC state data formatter is enabled with in memory distributed cache.

I've had a few cases in support where the Oidc state data formatter has been enabled but the distributed cache is the default memory based version. IdentityServer (unfortunately?) adds the in memory implementation as a fallback so it is easy to miss this. A warning message similar to the ones we have for in memory persisted grants should help.

### DIFF
--- a/src/IdentityServer/Configuration/DependencyInjection/ConfigureOpenIdConnectOptions.cs
+++ b/src/IdentityServer/Configuration/DependencyInjection/ConfigureOpenIdConnectOptions.cs
@@ -4,6 +4,9 @@
 
 using Duende.IdentityServer.Infrastructure;
 using Microsoft.AspNetCore.Authentication.OpenIdConnect;
+using Microsoft.Extensions.Caching.Distributed;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using System;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


src/IdentityServer/Configuration/DependencyInjection/ConfigureOpenIdConnectOptions.cs
```diff
@@ -21,12 +24,26 @@ public ConfigureOpenIdConnectOptions(string[] schemes, IServiceProvider serviceP
...
     {
         // no schemes means configure them all
         if (_schemes.Length == 0 || _schemes.Contains(name))
         {
             options.StateDataFormat = new DistributedCacheStateDataFormatter(_serviceProvider, name);
         }
+
+        var distributedCacheService = _serviceProvider.GetRequiredService<IDistributedCache>();
+
+        if (distributedCacheService is MemoryDistributedCache && !warnedInMemory)
```
josephdecock, line 39: We could do the `!warnedInMrmory` check before resolving the cache to avoid doing that work more than once.

AndersAbel:
Marking this as draft to keep it on hold until build issues in releases/7.0.x have been fixed.

josephdecock:
The 7.0 release should be fixed now, yet the checks still failed? That was so weird that I rebased this branch to try again...

josephdecock:
Okay, not sure why a rebase was necessary, but it looks like the build works now. @AndersAbel - do we need this in a 7.0 patch release?

AndersAbel:
@josephdecock I don't think we need a release just for this. I've had a couple of support cases caused by people running the OIDC Formatter with the in memory distributed cache. This will reduce the risk for people hitting that, but it can go with the next release.

PR #1559, opened 2024-05-24
Update dependencies for 6.3.x

AndersAbel:
Fixes [Support/#1266](https://github.com/DuendeSoftware/Support/issues/1266)


josephdecock (approved):

josephdecock (approved):
Wilson version 6.35.0 is the same as the version depended on by the asp.net OIDC handler, and upgrades our transitive dependencies to not use a package with a known vulnerability. It also matches what we depend on in AccessTokenManagement v2.

brockallen (approved):

PR #1570, opened 2024-06-04
Update ASP.NET core depedencies

josephdecock:
Update framework dependencies to avoid taking a transitive dependency on a vulnerable version of System.IdentityModel.Tokens.Jwt. Even though we specify wilson version, we aren't explicitly using the wilson version in most of the nuget packages we produce.


AndersAbel (approved):
This looks like the right versions.

PR #1573, opened 2024-06-09
Use SHA*.HashData() one-shot methods

martincostello:
Use `SHA*.HashData()` one-shot methods to avoid allocating `SHA*` objects.

This is something I just happened to notice while I was debugging an issue in something using IdentityServer in a sample and I stepped into the decompiled code.

The only usages of `SHA*.Create()` left after this change are these:

https://github.com/DuendeSoftware/IdentityServer/blob/e9860c6488f90e8fbc11a4452b9dd111dbfae933/src/IdentityServer/Configuration/CryptoHelper.cs#L84-L100

As this is a public method I left it as-is, but refactored the only caller within the codebase itself to use the one-shot methods and not depend on it anymore.

### DIFF
--- a/src/IdentityServer/Configuration/CryptoHelper.cs
+++ b/src/IdentityServer/Configuration/CryptoHelper.cs
@@ -69,16 +69,23 @@ public static class CryptoHelper
     /// <returns></returns>
     public static string CreateHashClaimValue(string value, string tokenSigningAlgorithm)
     {
-        using (var sha = GetHashAlgorithmForSigningAlgorithm(tokenSigningAlgorithm))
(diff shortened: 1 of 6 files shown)


josephdecock:
Okay thanks. I want to think a bit more about GetHashAlgorithmForSigningAlgorithm (maybe we deprecate it), but this great! For now I'll leave this open as a reminder.

josephdecock:
Thanks again for this contribution! I want to get this merged in, so I rebased this to get the latest from main (needed for the build). Now I've recreated this PR here: https://github.com/DuendeSoftware/IdentityServer/pull/1640, and am closing this.

PR #1583, opened 2024-08-16
use static equals method to avoid exception

testfirstcoder:
`user.Password` could be null.

### DIFF
--- a/src/IdentityServer/Test/TestUserStore.cs
+++ b/src/IdentityServer/Test/TestUserStore.cs
@@ -44,7 +44,7 @@ public class TestUserStore
                 return true;
             }
 
-            return user.Password.Equals(password);
+            return Equals(user.Password, password);
         }
 
         return false;


josephdecock:
Is it actually possible to create a TestUser instance without a password, or to submit the QuickStart ui's login page without entering a password? If so, maybe we should prevent that. Ultimately the TestUserStore isn't meant for production, so we should just make the developer experience as good as possible.

testfirstcoder:
The [previous line](https://github.com/DuendeSoftware/IdentityServer/blob/main/src/IdentityServer/Test/TestUserStore.cs#L42) checks explicit the possible null value for user password. 

> if (string.IsNullOrWhiteSpace(user.Password)

Right. The type ist for testing purpose only. 

Nevertheless we could use [Object.Equals](https://learn.microsoft.com/en-us/dotnet/api/system.object.equals) or [String.Equals](https://learn.microsoft.com/en-us/dotnet/api/system.string.equals) methods to make a safer comparision and to avoid a possible `System.NullReferenceException`.

josephdecock (approved):

josephdecock:
Okay, I had some thought of possibly looking at nullability annotations to make this impossible, but on second thought I'm just going to go ahead and merge it and not worry about the annotations.

PR #1586, opened 2024-08-20
Avoid multiple check adding clientid

testfirstcoder:
Reuse already existing check in the [AuthenticationPropertiesExtensions.AddClientId](https://github.com/DuendeSoftware/IdentityServer/blob/main/src/IdentityServer/Extensions/AuthenticationPropertiesExtensions.cs#L101) method.


josephdecock (approved):
Thanks!

PR #1592, opened 2024-09-16
Joe/7.0.6 to main

josephdecock:
Merge 7.0.6 forward to main


josephdecock:
This merges all of the 7.0.x release branch into main, but most of the commits end up being an empty diff because the changes were in both places already. All that is actually here is the fix for the is local url validation.

brockallen (approved):

PR #1595, opened 2024-09-18
Avoid unhandled argument null exception for refresh_token grant in token validator when scopes missing

Bartizan:

On call to /connect/token endpoint, Identity Server returns an HTML page due to an internal error instead of a response in JSON format.
[The related issue](https://github.com/DuendeSoftware/Support/issues/1409) is created.


josephdecock:
Thanks for the PR! I wonder if this is the right place to solve the problem? It seems like creating a refresh token that can't be used ought to be invalid, so maybe we shouldn't even allow the refresh token to be created in the first place.

josephdecock:
Closing in favor of https://github.com/DuendeSoftware/IdentityServer/issues/1611 (see also https://github.com/DuendeSoftware/Support/issues/1409)

PR #1597, opened 2024-09-18
Build updates in preparation for .net 9

josephdecock:
This PR prepares the repo for .net 9.

_Note that this only affects our development hosts and test projects, not any published package's dependency. All our current packages have no transitive security vulnerabilities as of now._

- Update Serilog to 8.0.2. This pulls in the right version of System.Text.Json in most cases.
- Pin versions of Azure.Identity (1.11.4), System.Formats.Asn1 (8.0.1), Microsoft.Data.SqlClient (5.1.6) and System.Text.Json. These are depended on by EF and Sql client packages, and there's no update to those packages available that wouldn't give us a vulnerable version. Hopefully someday those packages will update such that this is no longer needed.
- Rename ClientModel. The new System.ClientModel library is a new dependency that we get from updating Azure.Identity, and its namespace name conflicts with our existing class name. There's no good way to resolve the ambiguity, short of just renaming the model. We need to do the same in the templates. (https://github.com/DuendeSoftware/IdentityServer.Templates/issues/56)
- Convert CsQuery to AngleSharp in test projects. These are libraries for doing DOM manipulation in C#. CsQuery is no longer maintained, and it has some messy dependencies. AngleSharp is an easy replacement.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -42,7 +42,7 @@
         <PackageReference Update="Microsoft.IdentityModel.JsonWebTokens" Version="$(WilsonVersion)"/>
         <PackageReference Update="Microsoft.IdentityModel.Protocols.OpenIdConnect" Version="$(WilsonVersion)"/>
         <PackageReference Update="System.IdentityModel.Tokens.Jwt" Version="$(WilsonVersion)"/>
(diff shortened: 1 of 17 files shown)


josephdecock:
Huh, looks like some weird issue with the migrations that I didn't run locally...

leastprivilege (approved):

PR #1605, opened 2024-09-22
Switched to the more performant System.Threading.Lock locks.

MarkCiliaVincenti:


josephdecock:
Hi, thanks for the PR. Did you encounter a performance problem in practice that this solves? Have you benchmarked this at all?

MarkCiliaVincenti:
> Hi, thanks for the PR. Did you encounter a performance problem in practice that this solves? Have you benchmarked this at all?

Hi, yes the benchmarks show around a 25% performance improvement between locking on an object and locking on System.Threading.Lock.

https://steven-giesel.com/blogPost/4cf6d48a-ec9d-4c68-961c-31fd8d8c1340

https://github.com/MarkCiliaVincenti/Backport.System.Threading.Lock/tree/master/Backport.System.Threading.Lock.Benchmarks

Directory.Build.props
```diff
@@ -20,6 +20,7 @@
         <DebugType Condition="'$(GITHUB_ACTIONS)' == 'true'">embedded</DebugType>
 
         <ContinuousIntegrationBuild Condition="'$(GITHUB_ACTIONS)' == 'true'">True</ContinuousIntegrationBuild>
+		<LangVersion>preview</LangVersion>
```
josephdecock, line 23: Is this necessary?

Directory.Build.targets
```diff
@@ -23,6 +23,7 @@
 
         <!--build related-->
         <PackageReference Include="MinVer" Version="4.3.0" PrivateAssets="All"/>
+		<PackageReference Update="Backport.System.Threading.Lock" Version="2.0.3"/>
```
josephdecock, line 26: I'd prefer not to introduce a new dependency. Unless there was a practical performance problem that this new locking construct fixed, I don't think backporting this to our .net 8 builds is really necessary. On the other hand, conditional compilation is also a pain. Maybe this change should wait for a future version where we are no longer supporting .net < 9.

Directory.Build.props
MarkCiliaVincenti: At the moment yes, to lock on a System.Threading.Lock instance you need to have LangVersion preview, which will probably not be a requirement once .NET 9.0 is officially released.

Directory.Build.targets
MarkCiliaVincenti: The backport library is conditionally enabled only for pre .NET 9.0. So with this you're able to lock as if you're only supporting net9. In fact even if you forget to remove the PackageReference once you drop support for net8, it will not be a problem since it is conditionally eliminated for net9 or later.

josephdecock: Oh, I didn't see at first that the backport is actually your open source work - seems like a great contribution to the dotnet community that you've made! 

However, we never take external dependencies except what's made by Microsoft (unavoidable since we are building for aspnet core). It just simplifies supply chain audits and compliance if we don't bring in anything else.

No offense meant at all to you! We definitely will revisit this for a future release.

MarkCiliaVincenti:
@josephdecock a new version of Backport.System.Threading.Lock has come out that acts as a source generator and basically dropping the DLL as a dependency. If this new development makes you reconsider, I will amend this PR accordingly.

PR #1606, opened 2024-09-23
Replace alice/bob/email.com domains with example domains

wenz:
Replace alice.com with alice.example.com, bob.com with bob.example.com, and email.com with example.com

Fixes #1419 (https://github.com/DuendeSoftware/Support/issues/1419)


josephdecock (approved):
Thanks!

PR #1625, opened 2024-11-17
Host and Client build Fixes

josephdecock:
- Update Otel libraries in test clients and hosts - Helps the manual tests build, as older otel libs have a transitive dependency on system.text.json with a security vulnerability.
- Use Duende.IdentiyModel in test clients


damianh (approved):
lfg

PR #1629, opened 2024-11-21
fix nullable in IsNullOrEmpty

SimonCropp:
better nullability hints for callers of IsNullOrEmpty and do a direct return of `!list.Any()`


damianh:
Hi @SimonCropp, don't think we'll merge this as is. Not because it's wrong, but we need to set `<Nullable>enable</Nullable>` on the code base as a whole.

PR #1636, opened 2024-11-21
fix incorrect as usage in GetAuthenticationTimeEpoch

SimonCropp:


josephdecock (approved):
This seems like a good change, especially given that if the type of identity is incorrect, id will be null, and the very next line will throw a null reference exception. Thanks!

PR #1638, opened 2024-11-22
Extension di method for token cleanup service

7645re:
We needed to inject a custom token cleaning service and persistent grant storage, there is a registration method for persistent grant storage, but not for the token cleaning service, why not add

### DIFF
--- a/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/Additional.cs
+++ b/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/Additional.cs
@@ -13,6 +13,7 @@ using System;
 using System.Net.Http;
 using Duende.IdentityServer;
 using Duende.IdentityServer.Configuration;
+using Duende.IdentityServer.EntityFramework;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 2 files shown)


josephdecock:
The token cleanup job is a feature of the EntityFramework based store implementation. If you're using a custom store, than cleaning up records in that custom store is something you're entirely responsible for.

Changing the `ITokenCleanupService` through DI is an interesting idea but it would require some more engineering work and would be a breaking change.  The benefit that I see is that it would allow you to get some of our basic logic (such as the fuzzing of start-times of the token cleanup work added in 7.0), but customize the store access to use your custom store.

The problem is that the token cleanup service is currently implemented entirely within Duende.IdentityServer.EntityFramework. We won't want to take a dependency on that package from the Duende.IdentityServer package, because we have lots of users who aren't using the EF stores. One could imagine moving the cleanup job out of the EF package, and then having a mechanism to register a cleanup service that could either be used by the EF package to register _its_ implementation, or by users who are customizing token cleanup, but that's likely a breaking change because we'd have to rearrange where all these types are declared and where the options are located.

So, thanks for the PR, but I am going to close it for the reasons above.

7645re:
no problem, thanks for the constructive response

PR #1639, opened 2024-11-23
Update OIDC clients to .NET 9

josephdecock:
Update our internal oidc clients to .NET 9, and give them some minor cleanups (disabling telemetry, fixing log levels, etc).

Notably, my work on PAR in the asp.net handler means all our OIDC test clients use PAR by default now. This means that we can discard the clients that specifically added PAR, as that is redundant
     - MvcPar is gone, in favor of MvcCode, etc
     - MvcJarPar is gone, in favor of MvcJarJwt, which is updated to use the new PAR event (the JAR it creates is pushed to the PAR endpoint, rather than sent to the authorize endpoint)

Also, I'm using the preview of Duende.AccessTokenManagement 3.1.


damianh (approved):
All good.

PR #1640, opened 2024-11-23
Recreating martincostello's PR (#1573), rebased onto main

josephdecock:
I rebased @martincostello's PR (#1573) onto main because we need the latest package updates to build. 

The original PR replaces the SHA256.Create() (and other SHA HashAlgorithm) calls with SHA*.HashData, which uses less memory by not allocating a SHA* object. It had left in place one public function that would actually return the SHA* object. We weren't using that anymore but it was public.

I deprecated that function, and added a new function that returns the appropriate SHA*.HashData function instead.


damianh (approved):
Looks good. Thanks @martincostello !

PR #1647, opened 2024-11-27
use some pattern matching

SimonCropp:


josephdecock:
@SimonCropp thanks for this PR. We're actually looking at tooling and automation around stylist preferences like this soon. Since we're going to be doing that in a broader and more automatic way soon, I'm going to close this PR - just to avoid churn on the subject. But thanks anyway!

SimonCropp:
no worries

PR #1649, opened 2024-11-27
remove some dead code

SimonCropp:


SimonCropp:
@damianh all this code is internal. so why the breaking label? what am i missing?

damianh:
Ah the class internal. Label removed.

josephdecock:
StringExtensions is copied between the core and EF projects, and for now I think the easiest thing is to keep them in sync. There's probably something clever that we can do to have only 1 version of the file in future. For now I'm just going to close this one, since it has merge conflicts too.

PR #1653, opened 2024-11-28
Move all files into /identity-server

damianh:
- In preparation to merge all commercial products into a single repository, move the IdentityServer codebase into `/identity-server`. 
- GitHubActions workflows paths adjusted accordingly.

**Important: Any code or remarks in your Pull Request are under the following terms:**

If You provide us with any comments, bug reports, feedback, enhancements, or modifications proposed or suggested by You for the Software, such Feedback is provided on a non-confidential basis (notwithstanding any notice to the contrary You may include in any accompanying communication), and Licensor shall have the right to use such Feedback at its discretion, including, but not limited to the incorporation of such suggested changes into the Software. You hereby grant Licensor a perpetual, irrevocable, transferable, sublicensable, nonexclusive license under all rights necessary to incorporate and use your Feedback for any purpose, including to make and sell any products and services.

(see [our license](https://duendesoftware.com/license/identityserver.pdf), section 7)

PR #1656, opened 2024-12-05
remove redundant Distinct

SimonCropp:


josephdecock:
I don't think we can prove with certainty that these calls to Distinct are always redundant. We also have merge conflicts here, and in the interest of trying to clear out all the PRs so that we can merge https://github.com/DuendeSoftware/IdentityServer/pull/1653, I'm just going to close this one.

SimonCropp:
@josephdecock they are being added to a hashset. so the distinct is redundant?

josephdecock:
The `AllowedAccessTokenSigningAlgorithms` that this is called on is an `ICollection<string>` with a default value that is a `HashSet`, so while it is true that in the vast majority of cases the call is redundant, it is possible that some caller creates a different ICollection and then messes up by duplicating values in it.

SimonCropp:
i dont think so. the loop does a dictionary contains check on the type

josephdecock:
If you add a unit test that verifies that the results are distinct, then we can merge. But please hold off until after The Big Merge. Thanks!

PR #1659, opened 2024-12-06
Fix client compilation issues

josephdecock:
- Duende.IdentityModel.OidcClient does not exist (yet), so switch back to IdentityModel.OidcClient
- Fix usage of DCR extensions to be compatible with latest IdentityModel


damianh (approved):
Think we need to get a version of that package out sooner rather than later.

PR #1660, opened 2024-12-06
Persist claim issuers in server side sessions

### LINKED ISSUE #1546: Claim Issuer not persisted when using Server Side Sessions
7

8

**Describe the bug**

When using Server Side Sessions, the issuer of the claim is not persisted, which means on rehydration the issuer is set to `LOCAL_AUTHORITY` which is the default for a claim.

This causes issues with the Sustainsys.SAML2 Single Logout as it uses the issuer from the claim in order to redirect to the identity provider. See https://github.com/Sustainsys/Saml2/blob/3780695449868fdeb708f0665eadd042a1a27153/Sustainsys.Saml2/WebSSO/LogOutCommand.cs#L168

**To Reproduce**

Setup identity server that uses Sustainsys.SAML2 as an external identity provider with single logout enabled and enable server side sessions.
Then attempt to logout.

**Expected behavior**

The claim should be rehydrated with the original issuer in order to allow single logout.

--

The question is, was the omission of persisting the issuer a design choice and this should be handled in a different manner?

josephdecock:
Resolves #1546

@AndersAbel this came up in context of a support issue you worked where someone couldn't logout using saml and sustainsys, so I've added you as a reviewer if you have time - is this what you were hoping for?

### DIFF
--- a/src/Storage/Stores/Serialization/ClaimLite.cs
+++ b/src/Storage/Stores/Serialization/ClaimLite.cs
@@ -11,4 +11,5 @@ public class ClaimLite
     public string Type { get; set; }
     public string Value { get; set; }
     public string ValueType { get; set; }
+    public string Issuer { get; set; }
 }
--- a/hosts/EntityFramework/Pages/Account/Login/Index.cshtml.cs
+++ b/hosts/EntityFramework/Pages/Account/Login/Index.cshtml.cs
@@ -11,7 +11,6 @@ using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.RazorPages;
-using static System.Runtime.InteropServices.JavaScript.JSType;
 
 namespace IdentityServerHost.Pages.Login;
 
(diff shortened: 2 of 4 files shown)


src/Storage/Stores/Serialization/ClaimLite.cs
```diff
@@ -11,4 +11,5 @@ public class ClaimLite
     public string Type { get; set; }
     public string Value { get; set; }
     public string ValueType { get; set; }
-}
\ No newline at end of file
+    public string Issuer { get; set; }
```
damianh, line 14: Any concerns about about existing valves being null ?

josephdecock: Nullable reference types aren't enabled in this file, but Issuer and ValueType both in practice can be null and we handle them similarly when we deserialize.

damianh (approved):
Looks good to me. But also wait for Anders

AndersAbel:
This should work, but it would be possible to improve the optimization for the default issuer.

All claims in a ClaimsIdentity from a remote authentication provider typically have the same issuer. For Saml2 it's the Entity Id of the Idp (e.g. "https://stubidp.sustainsys.com/Metadata"). For OIDC it's the issuer/authority (e.g. "https://demo.duendesoftware.com").

For storage in a DB I don't think it's an issue to store the same string for each claim. As far as I remember this code would not be used to store anything in a cookie. If that was the case. I would have tried to optimize further to keep the issuer string once somewhere.

PR #1661, opened 2024-12-07
Fix static signing in hosts

josephdecock:
Constructing an `X509Certificate2` instance for a p12 file is problematic. The private key gets stored in a temp file, which is discarded when the `X509Certificate2` instance is disposed. That causes problems for us, as the public key material is in our store, but the private key no longer exists on disk.

Instead of instantiating, we should store the certificate in the OS store, and load the cert that we need from there.


hosts/Configuration/IdentityServerExtensions.cs
```diff
@@ -35,7 +35,6 @@ internal static WebApplicationBuilder ConfigureIdentityServer(this WebApplicatio
             .AddInMemoryIdentityResources(Resources.IdentityResources)
             .AddInMemoryApiScopes(Resources.ApiScopes)
             .AddInMemoryApiResources(Resources.ApiResources)
-            //.AddStaticSigningCredential()
```
josephdecock, line 38: No need to fix this in both main and configuration hosts. The configuration host doesn't need this.

damianh (approved):

PR #1665, opened 2024-12-09
elide some asyncs

SimonCropp:
I look through the existing codebase and it seems the preference is to elide asyncs where possible. to here are some that have been missed

### DIFF
--- a/src/AspNetIdentity/ProfileService.cs
+++ b/src/AspNetIdentity/ProfileService.cs
@@ -68,12 +68,12 @@ public class ProfileService<TUser> : IProfileService
     /// </summary>
     /// <param name="context">The context.</param>
     /// <returns></returns>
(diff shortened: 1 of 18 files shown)


josephdecock:
We don't have a clear decision on style for this. Your other PR where you pointed out that sometimes eliding async affects timing of activities makes me think we shouldn't change this without really carefully reviewing each change to make sure we're not doing something like that. But since we don't even have a consensus on which way we would go, I'm just going to close this - thanks anyway!

SimonCropp:
> shouldn't change this without really carefully reviewing each change to make sure we're not doing something like that

i was very careful. i reviewed each case to ensure it would not have those side effects

josephdecock:
Oh I'm sure you were careful, I didn't mean to imply otherwise. I was just noting that the code review would need to be quite careful as well - but since we don't have a commitment to this particular style, I don't think we need to make this change right now.

PR #1670, opened 2024-12-10
fix(error-redirect): use query-safe fragment as recommended in the latest OAuth 2 security best practices

aomader:
This PR adjusts the fragment that is added to error redirect responses, which is supposed to prevent browsers from reattaching existing ones.

The previous fragment `#_=_` is an invalid query selector and can not be used as is on the receiving end, hence it has been adjusted to the recommended `#_`.

Additionally, I adjusted the link to the respective latest active internet draft.

This looks like a minor thing, but one can run into issues caused by the browsers behavior of carrying along fragments.

### DIFF
--- a/identity-server/src/IdentityServer/Endpoints/Results/AuthorizeResult.cs
+++ b/identity-server/src/IdentityServer/Endpoints/Results/AuthorizeResult.cs
@@ -177,8 +177,8 @@ public class AuthorizeHttpWriter : IHttpResponseWriter<AuthorizeResult>
 
         if (response.IsError && !uri.Contains('#'))
         {
-            // https://tools.ietf.org/html/draft-bradley-oauth-open-redirector-00
-            uri += "#_=_";
+            // https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics-29#section-4.1.3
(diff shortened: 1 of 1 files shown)


aomader:
I guess it's a combination of the behavior of the browser to keep the fragment until it reaches the final address and that we do not know what the final address will be. The chain of redirects can be fairly long, with the fragment being added to some URL that eventually assumes something about fragments.

More precisely, we noticed that a client reacts on errors be performing redirects that end up on a page where fragments are assumed to be valid query selectors. In context of `text/html` as media type, I guess that this is a valid assumption, since they are document fragment identifiers.

Given that the recommendation was changed from a **non**-query-selector-safe fragment to a query-selector-safe one, and that following that recommendation is a non-breaking change, I would argue that this would be a generally beneficial adjustment.

aomader:
@josephdecock Any thoughts? We are seeing users being affected by this.

aomader:
In the open source context generally yes, but we are a paying customer (Enterprise License) and I am currently not having enough capacities to do this in my free time.

josephdecock:
That's absolutely fair - we'll add the test coverage and get this included in the 7.2 release.

PR #1674, opened 2024-12-20
cleanup Resources

SimonCropp:
* fix the nullability of the inputs in one of the constructors
 * avoid the multiple enumerations of the inputs in 2 of the constructors
 * avoid 3x HashSet allocations for 2 of the constructors

Question: should the 3 properties be changes to be ISet instead of ICollection?


josephdecock (approved):
This is looks good, thanks! For now let's keep the types unchanged, as technically switching away from IEnumerable is a breaking change.

PR #1675, opened 2024-12-20
ThrowIfNullOrWhiteSpace for type in IdentityProvider

SimonCropp:
I assume empty and whitespace is invalid for type?


josephdecock (approved):
This looks fine to me - I can't picture anyone ever correctly configuring an IdentityProvider with a type that was just white space.

PR #1676, opened 2024-12-20
remove one of the IsMissing extensions and move to ArgumentException.ThrowIfNullOrWhiteSpace

SimonCropp:
also removes 2x redundant ArgumentException check in CreatePrincipal since SubjectId is immutable and guarded in the constructor


josephdecock:
Thanks for this PR, but I think we're going to stick with the IsMissing method, at least for now. We like the symmetry between IsMissing and IsPresent, and having only one of them feels kind of incomplete.

PR #1677, opened 2024-12-20
enable nullable in TokenValidator

SimonCropp:
and fix 2x null reference exceptions

### DIFF
--- a/identity-server/src/IdentityServer/Validation/Default/TokenValidator.cs
+++ b/identity-server/src/IdentityServer/Validation/Default/TokenValidator.cs
@@ -2,6 +2,7 @@
 // See LICENSE in the project root for license information.
 
 
+#nullable enable
 using Duende.IdentityModel;
 using Duende.IdentityServer.Extensions;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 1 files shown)


identity-server/src/IdentityServer/Validation/Default/TokenValidator.cs
```diff
@@ -197,18 +198,18 @@ public async Task<TokenValidationResult> ValidateAccessTokenAsync(string token,
...
+            var principal = Principal.Create("tokenvalidator", result.Claims!.ToArray());
 
             if (result.ReferenceTokenId.IsPresent())
             {
                 principal.Identities.First()
                     .AddClaim(new Claim(JwtClaimTypes.ReferenceTokenId, result.ReferenceTokenId));
             }
 
-            var isActiveCtx = new IsActiveContext(principal, result.Client,
+            var isActiveCtx = new IsActiveContext(principal, result.Client!,
```
josephdecock, line 212: I don't think we need this !, because the IsActiveContext ctor is going to check for nulls.

josephdecock:
I'm going to close this because technically it would be a breaking change, since the null checks will cause new compiler errors for some users. It's something we should do, but probably we want to do these kinds of changes in 8.0

PR #1678, opened 2024-12-20
fix null ref in UserInfoRequestValidator

SimonCropp:


identity-server/src/IdentityServer/Validation/Default/UserInfoRequestValidator.cs
```diff
@@ -114,7 +114,7 @@ public async Task<UserInfoRequestValidationResult> ValidateRequestAsync(string a
         }
 
         // make sure user is still active
-        var isActiveContext = new IsActiveContext(subject, tokenResult.Client, IdentityServerConstants.ProfileIsActiveCallers.UserInfoRequestValidation);
+        var isActiveContext = new IsActiveContext(subject, tokenResult.Client!, IdentityServerConstants.ProfileIsActiveCallers.UserInfoRequestValidation);
```
josephdecock, line 117: I don't think we need the ! here - `IsActiveContext` handles nulls already.

SimonCropp: IsActiveContext handles nulls a runtime. this PR suppresses a null inconsistency a build time

josephdecock: The whole situation is kind of goofy.

- I generally only use a ! to mean that I am convinced that a value is impossible to be null, even though the static analysis can't figure it out
- The code we're calling to get the tokenResult is complex enough that I would worry about missing some edge case where it actually is possible to get a null
-  The IsActiveContext constructor parameters aren't nullable, but then it immediately checks them for null

So I think the real problem here is that IsActiveContext parameters should be changed to allow nulls at compile time. That would better reflect its actual behavior and the way we are calling it.

josephdecock: Hmm, I tried the API the way I suggested, and it was worse 😢 

Making the ctor parameters nullable only to throw if they are null is misleading and bad.

On reflection, we actually do know that if we have a valid token, there has to be a client associated with that token. So, turns out the ! is the simplest thing after all.

josephdecock:
I'm trying to reset to a previous commit (basically removing my suggestion that I decided I don't like), and github isn't letting me. There's probably some nicer way to do it, but I'm just going to close this PR and reopen a new one.

SimonCropp:
@josephdecock for my PRs, if there are ever git issues. dont hesitate to close it and ask me to resubmit

PR #1712, opened 2025-01-08
Fix incorrect 'for' attribute in userCode inputs

khalidabuhakmeh:
Replaced the incorrect 'for' attribute with 'id' for the userCode input fields across multiple Index.cshtml files to ensure proper HTML semantics and accessibility. This change aligns the input fields with their respective labels.


maartenba (approved):

josephdecock (changes requested):
Let's make the same changes in the templates subdirectory.

josephdecock (approved):

PR #1718, opened 2025-01-08
IdentityServer (hosts and templates) - summarize license usage on shutdown

josephdecock:
~~This PR is stacked on top of https://github.com/DuendeSoftware/products/pull/1717. Please merge it first.~~ _Done_

This PR copies the `LicenseUsageSummary` code from the main host into all the other hosts and into the templates.

I'm doing it in main rather than the release branch because
- hosts aren't released
- templates aren't synced with a release branch (at least not yet)

For this to compile in the templates, we need the latest release, so this depends on 7.1.0-rc.2. We MUST update to 7.1.0 when it hits GA before releasing the templates.

This also requires us to update the template files to refer to Duende.IdentityModel instead of IdentityModel.

### DIFF
--- a/identity-server/hosts/main/Program.cs
+++ b/identity-server/hosts/main/Program.cs
@@ -38,12 +38,17 @@ try
         .ConfigureServices()
         .ConfigurePipeline();
 
-    var usage = app.Services.GetRequiredService<LicenseUsageSummary>();
+    if (app.Environment.IsDevelopment())
+    {
+        app.Lifetime.ApplicationStopping.Register(() =>
+        {
+            var usage = app.Services.GetRequiredService<LicenseUsageSummary>();
+            Console.Write(Summary(usage));
+            Console.ReadKey();
+        });
+    }
 
     app.Run();
-
-    Console.Write(Summary(usage));
-    Console.ReadKey();
 }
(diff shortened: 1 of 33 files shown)


identity-server/hosts/main/Program.cs
```diff
@@ -38,12 +38,19 @@
...
         .ConfigurePipeline();
 
-    var usage = app.Services.GetRequiredService<LicenseUsageSummary>();
+    if (app.Environment.IsDevelopment())
+    {
+        app.Lifetime.ApplicationStopping.Register(() =>
+        {
+            var usage = app.Services.GetRequiredService<LicenseUsageSummary>();
+            Console.Write(Summary(usage));
+            {
```
khalidabuhakmeh, line 47: I'm not sure if this is a merge mistake, but there seems to be an extra `{}` scope here.

josephdecock: Thanks for catching - fixed!

khalidabuhakmeh: This is the same in the other PR I just approved. Not a "huge" deal but you may want to update their too.

khalidabuhakmeh (approved):

PR #1728, opened 2025-01-23
Made tracing class public to allow activity source extensions

marcindruzgala:

Currently `Tracing` class is marked as `internal` which stops us from extending `ActivitySource` declared in that class. Why would we want to extend it?

Duende Server currently adds few predefined values to traces produced by it and does it at different stages e.g. top level span for `connect/authorize` doesn't include `clientId` label. We would like to enrich all calls by `clientId` (wherever applicable of course) and with other labels depending on scenario.

How would that work? Example below

```csharp
Tracing.DuendeActivitySource.AddActivityListener(new ActivityListener
{
    ShouldListenTo = _ => true,
    ActivityStarted = activity =>
    {
        if (activity.Parent.TryGetCorrelationId(out var correlationId))
        {
            activity.SetTag(CorrelationIdKey, correlationId);
        }
    },
    ActivityStopped = activity =>
    {
       // enrich span with all key/value in Baggage
        foreach (var (key, value) in Baggage.Current)
        {
            activity.SetTag(key, value);
        }
    },
});
```

### DIFF
--- a/identity-server/src/Telemetry/Tracing.cs
+++ b/identity-server/src/Telemetry/Tracing.cs
@@ -9,7 +9,7 @@ namespace Duende.IdentityServer;
 /// <summary>
 /// Constants for tracing
 /// </summary>
-internal static class Tracing
+public static class Tracing
 {
     private static readonly Version AssemblyVersion = typeof(Tracing).Assembly.GetName().Version;
 


josephdecock (approved):
Sorry to keep this waiting a long time, but yes - we can do this! Thanks for your contribution!

PR #1729, opened 2025-01-24
Cleanup for the BFF project

Erwinvandervalk:
This PR contains several cleanups, including introducing of a editorconfig file. We'll align further on a standardized editorconfig in the near future. Also, it aligns the internal variable names with the editorconfig. 

It should not introduce any breaking changes.


.vs/slnx.sqlite
damianh, line 1: The `.vs` folder should probably be `gitignored`

bff/samples/Apis/Api.DPoP/Extensions.cs
```diff
@@ -33,7 +34,7 @@ public static WebApplication ConfigureServices(this WebApplicationBuilder builde
             });
 
         // layers DPoP onto the "token" scheme above
-        services.ConfigureDPoPTokensForScheme("token");
+        Duende.AspNetCore.Authentication.JwtBearer.DPoP.DPoPServiceCollectionExtensions.ConfigureDPoPTokensForScheme(services, "token");
```
josephdecock, line 37: Could this just be a using?

damianh: Looks like the call to an extension method was changed to it's static invocation form.

bff/samples/Hosts.Tests/TestInfra/AppHostFixture.cs
```diff
@@ -35,7 +31,7 @@ public IDisposable ConnectLogger(WriteTestOutput output)
 
     public bool UsingAlreadyRunningInstance { get; private set; }
     public string StartupLogs => _startupLogs.ToString() ?? string.Empty;
-    protected DistributedApplication? App = null!;
+    protected DistributedApplication? App;
     private Logger _logger = null!;
```
damianh, line 35: Take the opportunity to arrange these? e.g. fields, ctor, properties, methods etc.

damianh (approved):
Approval pending comment resolution.

PR #1730, opened 2025-01-24
Fix XML documentation syntax errors

maartenba:
Missing parameters and some syntax error fixes


josephdecock (changes requested):
This looks good, and we should make corresponding changes in https://github.com/DuendeSoftware/products/tree/main/templates.

josephdecock (approved):

PR #1733, opened 2025-01-24
fix(error-redirect): use query-safe fragment as recommended in the latest OAuth 2 security best practices

bhazen:
This is a PR based off the work done in [PR 1670](https://github.com/DuendeSoftware/products/pull/1670), but with test coverage added.

From the original author of the original PR:
This PR adjusts the fragment that is added to error redirect responses, which is supposed to prevent browsers from reattaching existing ones.

The previous fragment `#_=_` is an invalid query selector and can not be used as is on the receiving end, hence it has been adjusted to the recommended `#_`.

Additionally, I adjusted the link to the respective latest active internet draft.

This looks like a minor thing, but one can run into issues caused by the browsers behavior of carrying along fragments.


josephdecock (approved):
Nice work!

PR #1735, opened 2025-01-27
tests write output via ITestOutputHelper

Erwinvandervalk:
This PR ensures that all tests now write the HTTP logs to the output.

### DIFF
--- a/bff/test/Duende.Bff.Tests/TestFramework/TestLoggerProvider.cs
+++ b/bff/test/Duende.Bff.Tests/TestFramework/TestLoggerProvider.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Duende Software. All rights reserved.
+// Copyright (c) Duende Software. All rights reserved.
 // See LICENSE in the project root for license information.
 
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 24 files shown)


bff/test/Duende.Bff.Tests/TestFramework/TestLoggerProvider.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Duende.Bff.Tests.TestFramework
 {
-    public class TestLoggerProvider : ILoggerProvider
+    public class TestLoggerProvider(WriteTestOutput? writeOutput = null, string? name = null) : ILoggerProvider
```
josephdecock, line 10: Can we fix this nullability compiler complain please?

josephdecock (approved):

bff/test/Duende.Bff.Tests/TestFramework/TestLoggerProvider.cs
Erwinvandervalk: I'll do a pass to enable nullable everywhere in a follow up pr, if that's ok with you? for now, the code is correct but the project nullability settings are not.

PR #1737, opened 2025-01-27
Add option for strict validation of assertion audiences

josephdecock:
Adds a new option that will enforce that the only private_key_jwt audience allowed is the issuer identifier, and that it is a string - not an array of a single value.

The audience of private key jwts has historically been somewhat open to implementation in various OAuth and OpenId specs. Some specs have said to use the issuer identifier, others the token endpoint or other endpoints, and some say to accept all. However, there is currently an effort from IETF and OpenID foundation to make audience requirements consistent across all specifications, which will be accomplished by only accepting the issuer identifier. The most recent FAPI (financial grade API) specification requires this strict validation, and other specifications are in the process of being updated as well.

### DIFF
--- a/identity-server/src/IdentityServer/Validation/Default/PrivateKeyJwtSecretValidator.cs
+++ b/identity-server/src/IdentityServer/Validation/Default/PrivateKeyJwtSecretValidator.cs
@@ -89,20 +89,7 @@ public class PrivateKeyJwtSecretValidator : ISecretValidator
             return fail;
         }
 
-        var validAudiences = new[]
-        {
-            // token endpoint URL
-            string.Concat(_urls.BaseUrl.EnsureTrailingSlash(), ProtocolRoutePaths.Token),
-            // issuer URL + token (legacy support)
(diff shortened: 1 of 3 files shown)


identity-server/src/IdentityServer/Validation/Default/PrivateKeyJwtSecretValidator.cs
```diff
@@ -111,15 +98,48 @@ await _issuerNameService.GetCurrentAsync(),
...
             ClockSkew = TimeSpan.FromMinutes(5)
         };
+        
+        if (_options.StrictClientAssertionAudienceValidation)
+        {
+            // New strict audience validation requires that the audience be the issuer identifier, disallows multiple
+            // audiences in an array, and even disallows wrapping even a single audience in an array 
+            tokenValidationParameters.AudienceValidator = (audiences, token, parameters) =>
+            {
+                var jwt = token as JsonWebToken;
```
josephdecock, line 114: Cast instead of as here.

bhazen (approved):

PR #1738, opened 2025-01-28
bff - Make user service methods async

Erwinvandervalk:
The DefaultUserService methods are now async. 

## Breaking change

If you have extended the DefaultUserService class and have overwritten either the .GetUserClaims() and/or .GetManagementClaims() method, you'll find now that these have now been replaced with async methods:

GetUserClaimAsync() and GetManagementClaimsAsync().


josephdecock (approved):
I'm surprised we didn't have to update any tests for this. I suppose our integration tests are excercising this code, without directly referencing the method names and signatures, so I guess it makes sense?

PR #1743, opened 2025-01-31
New build pipelines

josephdecock:
Create build pipelines similar to what we do in the foss repo.


.github/workflow-gen/Program.cs
```diff
@@ -0,0 +1,301 @@
...
+using static GitHubContexts;
+
+var contexts = Instance;
+Component[] components = [
+    new("ignore-this",
+        ["IgnoreThis"],
+        ["IgnoreThis.Tests"],
+        "it"),
+
+    new("bff",
```
Erwinvandervalk, line 14: To be able to run playwright tests, i'll need to be able to install custom tools on the build agent for the BFF.. any thoughts on how we could tackle this in this approach?

Erwinvandervalk (approved):

.github/workflow-gen/Program.cs
damianh: The purpose of workflow-gen is composable and reusable steps. While reuse of workflows works well in Foss repo because it's all exactly the same there is no hard rule - products are likely to deviate. We'll do what's necessary to minimise friction (which may even mean not using workflow-gen ultimately).

damianh (approved):

Erwinvandervalk:
Doing a squash merge because the commit journey doesn't appear too valuable to retain.

PR #1745, opened 2025-02-03
move structured logging into aspire

Erwinvandervalk:
Using serilog in the samples means the logs don't go into aspire's structured logging, which kind of defeats the purpose. 
So, I'm reverting the use of serilog in the samples. 

### DIFF
--- a/bff/samples/Apis/Api.DPoP/Api.DPoP.csproj
+++ b/bff/samples/Apis/Api.DPoP/Api.DPoP.csproj
@@ -7,7 +7,6 @@
     <ItemGroup>
         <PackageReference Include="Duende.IdentityModel" />
         <PackageReference Include="Duende.AspNetCore.Authentication.JwtBearer" />
-        <PackageReference Include="Serilog.AspNetCore" />
     </ItemGroup>
 
 	<ItemGroup>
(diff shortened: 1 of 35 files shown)


josephdecock:
We have a lot samples that are using Serilog. I think it would make sense to either keep serilog here or remove it more broadly. I think we can keep using Serilog with Aspire by using the Serilog.Sinks.OpenTelemetry package. See for instance [this blog](https://blog.nashtechglobal.com/distributed-tracing-with-serilog-and-opentelemetry/)

josephdecock (commented):
We have a lot samples that are using Serilog. I think it would make sense to either keep serilog here or remove it more broadly. I think we can keep using Serilog with Aspire by using the Serilog.Sinks.OpenTelemetry package. See for instance [this blog](https://blog.nashtechglobal.com/distributed-tracing-with-serilog-and-opentelemetry/)

damianh:
These hosts are at used dev & test(s) time and run via aspire with structured logging going to aspire dashboard. Unless I'm missing something, I'm not clear on what specific value `Serilog` is bringing to us here that helps us develop BFF and these hosts.

The actual samples that are part of public docs are a different concern and remain unchanged with respect to `Serilog` usage in the context of this PR.

josephdecock (approved):

PR #1756, opened 2025-02-04
Refactored the use of ClaimLite and ClaimsRecord into a single type. 

Erwinvandervalk:
Refactor The use of ClaimLite and ClaimsRecord into a single type. 

Note, ClaimsLite now has to be public as it's part of the signature of the DefaultUserService. 


bff/src/Duende.Bff.Blazor.Client/ClaimRecord.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System.Text.Json.Serialization;
+
+namespace Duende.Bff;
+
+/// <summary>
+/// Serialization friendly claim.
+///
+/// Note, this is a copy of the ClaimRecord class from Duende.Bff, but since we can't create a reference to it, we need to copy it here.
+/// We also can't link to it (as we do with the extensions) because we had to make it public. 
+/// </summary>
```
josephdecock, line 10: Comment seems out of date, since this isn't public anymore. Does that mean we can link to a shared file?

bff/test/Duende.Bff.Tests/TestHosts/BffHost.cs
```diff
@@ -156,7 +156,7 @@ private void Configure(IApplicationBuilder app)
                         context.Request.Path.Value ?? "/",
                         context.User.FindFirst("sub")?.Value,
                         context.User.FindFirst("client_id")?.Value,
-                        context.User.Claims.Select(x => new ClaimRecord(x.Type, x.Value)).ToArray())
+                        context.User.Claims.Select(x => new TestFramework.ClaimRecord(x.Type, x.Value)).ToArray())
```
josephdecock, line 159: Does there need to be a separate ClaimRecord in TestFramework? Can it get in on the shared goodness?

Erwinvandervalk: Actually it can and it should and it is now :)

bff/src/Duende.Bff.Blazor.Client/ClaimRecord.cs
Erwinvandervalk: as discussed, i changed the comment.

PR #1757, opened 2025-02-04
Persistent Cookies, Server Side Sessions, and Session Lifetime Coordination Fix

bhazen:
We had a bug report from a customer where when using a combination of persistent cookies and server side sessions with session lifetime coordination enabled, the persistent cookie would expire due to how the issued and expire dates of the session are moved forward in conjunction with how the .NET cookie auth handler triggers renewal of cookies. These changes address that specific scenario.


identity-server/src/IdentityServer/Stores/Default/ServerSideSessionCookieHelpers.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Duende Software. All rights reserved.
+// See LICENSE in the project root for license information.
+
+
+using Microsoft.AspNetCore.Authentication.Cookies;
+
+namespace Duende.IdentityServer.Stores.Default;
+
+internal static class ServerSideSessionCookieHelpers
```
josephdecock, line 9: I don't love calling things "Helpers" - Maybe `ServerSideSessionCookieEvents`?

josephdecock (approved):

PR #1764, opened 2025-02-05
build - build with the correct security context

Erwinvandervalk:
This PR changes the build so that a build runs only once, and in the right security context, depending on their origins. 

_Previously, the build used pull_request_target. Effectively, this means that the entire context (including which code to build) was pointing to the target of the PR (likely main). This means it wasn't building anything in the PR, but actually building main (not very useful in a PR workflow)_ 

### How does it work?
1. Builds / pr's from contributors are run trusted and can access security tokens needed for signing. If you create a branch in the DuendeSoftware repo, then you'll automatically get a CI build for this. You can (must) create a PR for this to get it merged. When you create a PR, the build script will detect that there is still a branch for it and will built it normally (skip the PR build).

2. Builds from PR's created from external contributors (which have to originate from a different fork) will NOT run with security tokens, so artifacts will not be signed / pushed. 

### Publishing artifacts somewhere from external contributors. 
/
If we ever want to to anything with the output of external pr's other than merge them into main (IE: deploy them somewhere), then we should:
a. let the build publish the artifacts to github. 
b. setup a different workflow that will run only if the after the previous build is complete AND if pr is labelled (which only contributors can do) This workflow can then publish the changes. 

### DIFF
--- a/.github/workflow-gen/Program.cs
+++ b/.github/workflow-gen/Program.cs
@@ -77,9 +77,7 @@ void GenerateCiWorkflow(Component component)
         job.StepPack(project);
     }
 
-    job.StepSign()
-        .OnlyIfOnBranch()
-        ;
+    job.StepSign();
 
     job.StepPush("MyGet", "https://www.myget.org/F/duende_identityserver/api/v2/package", "MYGET");
 
@@ -193,7 +191,7 @@ public static class StepExtensions
     /// Only run this if the build is triggered on a branch IN the same repo
     /// this means it's from a trusted contributor.
     /// </summary>
-    public static Step OnlyIfOnBranch(this Step step)
+    public static Step IfGithubEventIsPush(this Step step)
         => step.If("github.event == 'push'");
 
     public static void StepTestAndReport(this Job job, string componentName, string testProject)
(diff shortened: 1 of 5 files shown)


.github/workflow-gen/Program.cs
```diff
@@ -180,9 +183,19 @@ public static void StepSetupDotNet(this Job job)
...
+    /// Only run this for a main build
+    /// </summary>
     public static Step IfRefMain(this Step step) 
         => step.If("github.ref == 'refs/heads/main'");
 
+    /// <summary>
+    /// Only run this if the build is triggered on a branch IN the same repo
+    /// this means it's from a trusted contributor.
+    /// </summary>
+    public static Step OnlyIfOnBranch(this Step step)
```
damianh, line 196: `IfEventIsPush` ?

.github/workflow-gen/Program.cs
```diff
@@ -76,7 +77,9 @@ void GenerateCiWorkflow(Component component)
         job.StepPack(project);
     }
 
-    job.StepSign();
+    job.StepSign()
+        .OnlyIfOnBranch()
+        ;
```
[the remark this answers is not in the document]
Erwinvandervalk: actually, the OnlyIfOnBranch() was a duplicate. So I removed it. And I put the semicolumn on the end ;)

.github/workflow-gen/Program.cs
Erwinvandervalk: Sure, I thought it would be clearer to indicate the intent, rather than the methodology, but this is more in line with the rest of the code. No strong preference, so I changed it.

damianh (approved):

PR #1766, opened 2025-02-05
introducing playwright test for blazor

Erwinvandervalk:
Introduces playwright tests (front-end) for the blazor samples.

- [x] Run tests against already running server
- [x] Make sure server is started
- [x] Make sure the build agent installs prerequisites.
- [x] Use service discovery
- [x] Introduce page models

### DIFF
--- a/.github/workflows/bff-ci.yml
+++ b/.github/workflows/bff-ci.yml
@@ -47,11 +47,10 @@ jobs:
       run: dotnet build bff.slnf -c Release
     - name: Dotnet devcerts
       run: dotnet dev-certs https --trust
+    - name: Install Playwright
+      run: pwsh test/Hosts.Tests/bin/Release/net9.0/playwright.ps1 install --with-deps
     - name: Test
       run: dotnet test bff.slnf -c Release --no-build --logger "console;verbosity=normal" --logger "trx;LogFileName=Tests.trx" --collect:"XPlat Code Coverage"
-    - name: test
-      run: 'ls '
-      working-directory: test
     - name: Test report
       if: github.event == 'push' && (success() || failure())
       uses: dorny/test-reporter@31a54ee7ebcacc03a09ea97a7e5465a47b84aea5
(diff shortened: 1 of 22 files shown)


damianh (approved):

.github/workflows/bff-ci.yml
```diff
@@ -47,13 +47,12 @@ jobs:
       run: dotnet build bff.sln -c Release
     - name: Dotnet devcerts
       run: dotnet dev-certs https --trust
+    - name: Install Playwright
+      run: pwsh test/Hosts.Tests/bin/Release/net9.0/playwright.ps1 install --with-deps
```
damianh, line 51: This is such [a horrible way to install](https://github.com/microsoft/playwright-dotnet/issues/2286) :(

.github/workflows/bff-ci.yml
```diff
@@ -47,13 +47,12 @@ jobs:
...
     - name: Test
       run: dotnet test bff.sln -c Release --no-build --logger "console;verbosity=normal" --logger "trx;LogFileName=Tests.trx" --collect:"XPlat Code Coverage"
```
damianh, line 53: Because Playwright is slow (slow to install because it downloads browsers and slow to run), to keep CI build times sane, I predict you'll want to run your playwright tests in a separate parallel job.

.github/workflows/bff-ci.yml
Erwinvandervalk: I totally agree. In the future I may also need nodejs for some scneairos.. there's a cleaner nodejs based install but I don't know if it works for the .net version of playwright.

.github/workflows/bff-ci.yml
Erwinvandervalk: I'll keep an eye out for it. For now, 3 minutes of buildtime is something can live with. 

The issue with carving out the playwright tests is that that path (testing including playwright) will take 3 minutes, so overall build time is not improved. To run the playwright tests, I have to build everything, run everything, then execute the playwright tests (which are quite fast).

PR #1775, opened 2025-02-07
Introducing one big solution and solution filters

Erwinvandervalk:
All solutions are removed and a single 'products' solution is introduced with product specific solution filters are introduced

Note, it might be a good idea to wait until .net 9.0.200 is released. Right now, you cannot just do 'dotnet build' in a solution that contains solution filters:
https://github.com/dotnet/msbuild/issues/9587
Apparently this is fixed in .net 9.0.200. 


josephdecock (approved):
This looks cool, and I think we can merge even without waiting for the next sdk release. My thinking is that everything works except for the implicit build of a slnf when that's the only file present. But that's ok in most cases, because we have multiple sln files already, so we already have to specify a sln every time (at least in IdentityServer).  Waiting to merge this until the sdk release seems like we will likely have merge conflicts as we continue to refine the builds, so I'd really like to merge quick.

PR #1777, opened 2025-02-07
Fix builds of identityserver client

josephdecock:
Fixes broken client builds


josephdecock:
Skipping review for this trivial change to builds of support tooling

PR #1781, opened 2025-02-09
Filter client credentials in ToOptimizedFullDictionary method

josephdecock:
Filter client authentication because it's possible for a PAR request to include client auth, and we don't want that to cause client auth parameters to get passed into the authorization parameters message store, where they might get logged or otherwise stored insecurely.

### DIFF
--- a/identity-server/test/IdentityServer.UnitTests/Extensions/ValidatedAuthorizeRequestExtensionsTests.cs
+++ b/identity-server/test/IdentityServer.UnitTests/Extensions/ValidatedAuthorizeRequestExtensionsTests.cs
@@ -47,4 +47,23 @@ public class ValidatedAuthorizeRequestExtensionsTests
 
         Assert.Equal(2, res[OidcConstants.AuthorizeRequest.Resource].Length);
     }
+
+    [Theory]
+    [Obsolete]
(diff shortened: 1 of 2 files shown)


identity-server/test/IdentityServer.UnitTests/Extensions/ValidatedAuthorizeRequestExtensionsTests.cs
```diff
@@ -47,4 +47,23 @@ public void ToOptimizedFullDictionary_should_return_dictionary_with_array_for_re
...
+        {
+            Raw = new System.Collections.Specialized.NameValueCollection
+            {
+                { filteredClaimType, "" },
+            }
+        };
+
+        var result = request.ToOptimizedFullDictionary();
+
+        Assert.Equal(0, result.Count);
```
bhazen, line 67: Thoughts on the comments from the GitHub warnings about this line? Is one more consistent with the rest of our codebase?

josephdecock: No - and actually XUnit asserts aren't commonly used. We usually use FluentAssertions or Shoudly. I used XUNit asserts here because that was consistent with this file and because the migration to Shouldly is in progress and didn't want to disrupt that.

PR #1784, opened 2025-02-10
Ensure session lifetime coordination is invoked when a session expires

bhazen:
Scenarios can exist where session lifetime coordination is enabled but a session is expired and removed prior to the cleanup job running and processing that expiration. These changes account for that scenario and implement a workaround to detect when the dotnet cookie handler removes a session and triggers the session coordination service's processing of an expired session to ensure things such as backchannel logout and removal of grants.

### DIFF
--- a/identity-server/src/IdentityServer/Extensions/HttpContextExtensions.cs
+++ b/identity-server/src/IdentityServer/Extensions/HttpContextExtensions.cs
@@ -29,6 +29,23 @@ public static class HttpContextExtensions
         return context.Items.ContainsKey(Constants.EnvironmentKeys.SignOutCalled);
     }
 
+    internal static void SetExpiredUserSession(this HttpContext context, UserSession userSession)
+    {
+        ArgumentNullException.ThrowIfNull(context);
+        context.Items[Constants.EnvironmentKeys.DetectedExpiredUserSession] = userSession;
+    }
+
+    internal static bool TryGetExpiredUserSession(this HttpContext context, out UserSession expiredUserSession)
+    {
+        expiredUserSession = null;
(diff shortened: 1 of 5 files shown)


josephdecock (approved):

identity-server/src/IdentityServer/Extensions/HttpContextExtensions.cs
```diff
@@ -29,6 +29,17 @@ internal static bool GetSignOutCalled(this HttpContext context)
         return context.Items.ContainsKey(Constants.EnvironmentKeys.SignOutCalled);
     }
 
+    internal static void SetExpiredUserSession(this HttpContext context, UserSession userSession)
+    {
+        ArgumentNullException.ThrowIfNull(context);
+        context.Items[Constants.EnvironmentKeys.DetectedExpiredUserSession] = userSession;
+    }
+
+    internal static bool TryGetExpiredUserSession(this HttpContext context, out object expiredUserSession)
```
brockallen, line 38: Why `object`? Does this have something to do with dependencies?

josephdecock: I think we're just returning the underlying value from the httpcontext's dictionary, where the values are Al just objects. When we actually use this later on, we cast it. I think it's probably better to move the cast inside this method for clarity.

bhazen: `object` had been used initially because we were returning the result of `context.Items.TryGetValue` and it required that type. Per @josephdecock's comment I updated the internals to cast the result and there and return the aggregated result of the try and that typing as the return value.

PR #1793, opened 2025-02-12
Re-organizing templates

Erwinvandervalk:
Reorganizes how we handle templates. 

1. Split up templates to separate templates for BFF and Identity Server
2. Modernized build / release infrastructure to use dotnet commands (dotnet pack, sign, etc)

# Breaking changes

This PR deprecates the Duende.IdentityServer.Templates in favor of Duende.Templates.IdentityServer and Duende.Templates.BFF.


nuget.config
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
```
damianh, line 2: I think we can delete this file.

identity-server/templates/build.ps1
```diff
@@ -0,0 +1,56 @@
+$TOOLS_DIR = Join-Path $PSScriptRoot "tools"
```
damianh, line 1: This file still needed?

bff/templates/nuget.config
damianh, line 1: This file still needed?

damianh (changes requested):
Looks good modern and arrangement wise.  Few things
-  The template readmes need updating. The IS one shouldn't mention BFF and vice versa. Title, and instructuctions with the new package names.
- the `build.ps1`, `nuget.config` and `global.json` could probably be removed?

bff/templates/nuget.config
Erwinvandervalk: no it's not, removed it.

identity-server/templates/build.ps1
Erwinvandervalk: no it's not, removed it.

nuget.config
Erwinvandervalk: agreed. removed it.

identity-server/templates/templates/IdentityServerAspNetIdentity/.template.config/template.json
```diff
@@ -2,12 +2,13 @@
...
+    "Duende",
     "IdentityServer"
   ],
   "name": "Duende IdentityServer with ASP.NET Core Identity",
-  "identity": "IdentityServer.AspNetIdentity",
-  "groupIdentity": "IdentityServer.AspNetIdentity",
-  "shortName": "isaspid",
+  "identity": "Duende.IdentityServer.AspNetIdentity",
+  "groupIdentity": "Duende.IdentityServer.AspNetIdentity",
+  "shortName": "is-asp-id",
```
josephdecock, line 11: The new naming convention that you've adopted is definitely better, but the downside is that old documentation and other content (3rd party blogs, pluralsight content, maybe even our own youtube tutorials) all refer to the old names. I think we should add the new names, but not break the old ones:
```suggestion
  "shortName": ["is-asp-id", "isaspid"],
```

bff/hosts/Hosts.AppHost/Program.cs
```diff
@@ -49,8 +50,13 @@
...
     .WithReference(apiDPop)
-    .WithReference(bffDPop)
-    ;
+    .WithReference(bffDPop);
+
+builder.AddProject<BffLocalApi>("templates-bff-local", launchProfileName: null)
+    .WithHttpsEndpoint(5300, name: "bff-local");
+
+builder.AddProject<BffRemoteApi>("templates-bff-remote", launchProfileName: null)
+    .WithHttpsEndpoint(5310, name: "bff-remote");
```
josephdecock, line 59: Really smart to add the templates themselves to aspire. We need to get aspire going in the same way on the IdentityServer side! Do we have automated tests that sanity check all the aspire projects? Should we add the templates to that?

bff/test/Hosts.Tests/PageModels/PerComponentPageModel.cs
```diff
@@ -1,14 +1,10 @@
-﻿using Microsoft.Playwright;
+﻿using Microsoft.AspNetCore.Mvc.RazorPages;
```
josephdecock, line 1: Is this meant to be part of the template changes?

josephdecock (changes requested):
Overall, looks awesome. My minor preference would be to not have a templates directory inside the templates directory. Can we move the files up a level and still have things neat and tidy?

josephdecock:
We should also look into creating a meta package that depends on all the product specific template packages. We could have that be the next version of the existing package. The goal is to 
- have one package to install to get all of our stuff
- have existing users get our updates by just doing "dotnet new update"
- not have the old and new templates installed side by side

bff/hosts/Hosts.AppHost/Program.cs
Erwinvandervalk: that is my plan, but one step at a time

bff/test/Hosts.Tests/PageModels/PerComponentPageModel.cs
Erwinvandervalk: yes, as a side effect of adding the templates to aspire, the aspire tests became unstable. I had to make some changes to make the more stable. 

Theoretically I could split up the adding of the templates to aspire in a separate PR, but I wanted to test if it all worked.

identity-server/templates/templates/IdentityServerAspNetIdentity/.template.config/template.json
Erwinvandervalk: Agreed, did this. Although we'll still break them via the names of the actual packages.

josephdecock (approved):

PR #1797, opened 2025-02-14
Prevent log warning when expected duplicate key constraint is violated. 

Erwinvandervalk:
There is a possible race condition when inserting the same session twice (for example when the session for two cloned browsertabs expires and wants to re-authenticate at the same time. 

Right now, the system logs an warning that the insert has failed, but doesn't indicate that this is actually not a problem. So this PR improves that error message. 

One challenge with implementing this PR is that our EF implementation is storage agnostic. This means we don't know what database provider our customer uses, so we do our best to detect if there actually IS a duplicate key error. It won't be bulletproof but better than what we had. 


bff/src/Bff.EntityFramework/Store/UserSessionStore.cs
```diff
@@ -55,7 +55,25 @@ public async Task CreateUserSessionAsync(UserSession session, CancellationToken
...
+            // SQL Server would send:  ---> Microsoft.Data.SqlClient.SqlException (0x80131904): Cannot insert duplicate key row in object 'Session.UserSessions' with unique index 'IX_UserSessions_ApplicationName_SessionId'. The duplicate key value is (<AppName>, <SessionIdValue>).
+            // Postgres would send:  ---> Npgsql.PostgresException (0x80004005): 23505: duplicate key value violates unique constraint "IX_UserSessions_ApplicationName_SessionId"
+            // MySQL would send:    ---> MySql.Data.MySqlClient.MySqlException (0x80004005): Duplicate entry '<AppName>-<SessionIdValue>' for key 'IX_UserSessions_ApplicationName_SessionId'
+            if (exception.Contains("UNIQUE", StringComparison.OrdinalIgnoreCase) || exception.Contains("IX_UserSessions_ApplicationName_SessionId"))
+            {
+                _logger.LogDebug(ex, "Failed to save session in database due to Unique constraint error. This is likely due to a race condition between deleting the old session and inserting a new session and can safely be ignored.");
```
josephdecock, line 71: Maybe we can make this message a bit more reassuring by explaining why this is a safe situation to ignore the exception.

bff/src/Bff.EntityFramework/Store/UserSessionStore.cs
```diff
@@ -55,7 +55,25 @@ public async Task CreateUserSessionAsync(UserSession session, CancellationToken
...
+            }
+            else
+            {
+                _logger.LogWarning(ex, "Exception creating new server-side session in database: {error}. Should this be a duplicate key error it's safe to ignore.", ex.Message);
```
josephdecock, line 75:
```suggestion
                _logger.LogWarning(ex, "Exception creating new server-side session in database: {error}. If this is a duplicate key error, it's safe to ignore.", ex.Message);
```

josephdecock (approved):
Looks good - some minor thoughts on the messages that you can use if you like.

PR #1808, opened 2025-02-25
Simplify Blazor WASM authentication state provider.

josephdecock:
Refactored the blazor wasm auth state provider. The idea is that the provider is in charge of "orchestration" - basically deciding when to poll and what to do with the auth state we might get from the server side as persistent component state. And, while we do have a couple of services to encapsulate fetching from the /bff/user endpoint and reading persistent component state, those don't need interfaces.


bff/test/Bff.Blazor.Client.UnitTests/FetchUserServiceTests.cs
```diff
@@ -0,0 +1,151 @@
...
+    public async Task GetUserAsync_returns_anonymous_when_http_request_fails()
+    {
+        var factory = TestMocks.MockHttpClientFactory("Internal Server Error", HttpStatusCode.InternalServerError);
+        var sut = new FetchUserService(factory, Substitute.For<ILogger<FetchUserService>>());
+
+        var errorResult = await sut.FetchUserAsync();
+        errorResult.Identity?.IsAuthenticated.ShouldBeFalse();
+    }
+
+    // [Fact]
```
Erwinvandervalk, line 49: Should this test be commented out?

josephdecock: No, deleted. That logic got moved into the auth state provider itself, and new tests there should cover the moved logic.

Erwinvandervalk (approved):

PR #1812, opened 2025-02-26
fix casing of bff.blazor packages

Erwinvandervalk:
the packages for BFF.Blazor and EF had a different casing than the other packages

This pr fixes that. 

image


josephdecock (approved):
We might consider moving the files to match the assembly names, but it's probably really annoying for you to fight the windows case insensitive file system and doesn't need to happen in this PR.

Erwinvandervalk:
> We might consider moving the files to match the assembly names, but it's probably really annoying for you to fight the windows case insensitive file system and doesn't need to happen in this PR.

Yes, I considered that and decided I would do this a bit later.

PR #1830, opened 2025-03-05
Add caching support for discovery documents

khalidabuhakmeh:
Introduced `PreviewFeaturesOptions` to enable and configure discovery document caching. Updated `DiscoveryEndpoint` to conditionally use caching based on these options. Refactored `DiscoveryDocumentResult` to pre-serialize the response for improved performance.

The serialization of the discovery document can be expensive. Under heavy load, it can spike resources, which may cause garbage collection, affecting the responsiveness of the IdentityServer instance.


identity-server/src/IdentityServer/Configuration/DependencyInjection/Options/PreviewFeaturesOptions.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    /// <summary>
+    /// Enables Caching of Discovery Document based on ResponseCaching Interval 
+    /// </summary>
+    public bool EnableDiscoveryDocumentCache { get; set; } = false;
+    
+    /// <summary>
+    /// DiscoveryDocument Cache in Seconds
+    /// </summary>
+    public TimeSpan DiscoveryDocumentCacheDuration{ get; set; } = TimeSpan.FromSeconds(60);
```
maartenba, line 16:
```suggestion
    /// <summary>
    /// DiscoveryDocument Cache Duration
    /// </summary>
    public TimeSpan DiscoveryDocumentCacheDuration{ get; set; } = TimeSpan.FromMinutes(1);
```

maartenba: (can be any timespan so dropped "seconds")

maartenba:
When merging will need: these merged as well:
https://github.com/DuendeSoftware/fenix/pull/5
https://github.com/DuendeSoftware/docs.duendesoftware.com/pull/611

identity-server/src/IdentityServer/Endpoints/DiscoveryEndpoint.cs
```diff
@@ -63,8 +65,31 @@ public async Task<IEndpointResult> ProcessAsync(HttpContext context)
...
-        var response = await _responseGenerator.CreateDiscoveryDocumentAsync(baseUrl, issuerUri);
+    
+#pragma warning disable DUENDEPREVIEW001
+        if (_options.Preview.EnableDiscoveryDocumentCache)
+#pragma warning restore DUENDEPREVIEW001
+        {
+            // we need a cache but since this is preview, we don't
+            // want to force folks to register IMemoryCache if they aren't caching.
+            // This requires we use the service locator pattern :(
+            var cache = context.RequestServices.GetService<IMemoryCache>();
```
maartenba, line 76: The IS code already has `IDistributedCache` registered by default I believe. Can you see if that's true and perhaps switch to using that?

identity-server/src/IdentityServer/Endpoints/DiscoveryEndpoint.cs
```diff
@@ -63,8 +65,31 @@ public async Task<IEndpointResult> ProcessAsync(HttpContext context)
...
+                // the cache key accounts for multi-tenancy in Enterprise instances 
+                return await cache.GetOrCreateAsync($"discoveryDocument/{baseUrl}/{issuerUri}", async entry =>
+                {
+                    entry.AbsoluteExpirationRelativeToNow = _options.Preview.DiscoveryDocumentCacheDuration;
+                    var response = await _responseGenerator.CreateDiscoveryDocumentAsync(baseUrl, issuerUri);
+                    return new DiscoveryDocumentResult(response, _options.Discovery.ResponseCacheInterval);
+                });
+            } 
 
+            _logger.LogWarning("Unable to create discovery document cache. Add IMemoryCache by calling services.AddMemoryCache().");
```
maartenba, line 89: Nice!

maartenba (changes requested):

identity-server/src/IdentityServer/Endpoints/Results/DiscoveryDocumentResult.cs
```diff
@@ -17,21 +17,18 @@ namespace Duende.IdentityServer.Endpoints.Results;
 /// <seealso cref="IEndpointResult" />
 public class DiscoveryDocumentResult : EndpointResult<DiscoveryDocumentResult>
 {
-    /// <summary>
-    /// Gets the entries.
-    /// </summary>
-    /// <value>
-    /// The entries.
-    /// </value>
-    public Dictionary<string, object> Entries { get; }
```
bhazen, line 26: Would this be a breaking change? This is a public non-sealed class someone could have inherited from to implement some customization.

khalidabuhakmeh: The class is public but it's used only by internal objects. I can leave the Entries dictionary on there but it doesn't do much other than hang around.

identity-server/src/IdentityServer/Endpoints/DiscoveryEndpoint.cs
khalidabuhakmeh: There are warnings around `IDisributedCache` in IdentityServer that hint that folks shouldn't use the `MemoryDistributedCache`, so I can't assume that an instance of `IDisributedCache` will be in memory.

https://github.com/DuendeSoftware/products/blob/511f1d5a7d54cc121464a0d79c57634c972722d3/identity-server/src/IdentityServer/Configuration/DependencyInjection/ConfigureOpenIdConnectOptions.cs#L39-L47

josephdecock (commented):
We'll include this change in 7.2, so let's rebase this onto releases/is/7.2.x

identity-server/src/IdentityServer/Configuration/DependencyInjection/Options/PreviewFeaturesOptions.cs
```diff
@@ -0,0 +1,21 @@
...
+/// <summary>
+/// Preview Features Options
+/// </summary>
+public class PreviewFeaturesOptions
+{
+    /// <summary>
+    /// Enables Caching of Discovery Document based on ResponseCaching Interval.
+    /// Important: Requires in memory caching be enabled (services.AddInMemoryCaching())
+    /// </summary>
+    [Experimental("DUENDEPREVIEW001", UrlFormat = "https://duende.link/previewfeatures?id={0}")]
```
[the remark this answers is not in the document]
khalidabuhakmeh: @maartenba has some documentation. It will likely just scroll to an anchor tag.

identity-server/src/IdentityServer/Endpoints/DiscoveryEndpoint.cs
josephdecock: This is a good point, and I like what you did checking that IMemoryCache resolves. Let's keep with IMemoryCache.

identity-server/src/IdentityServer/Endpoints/Results/DiscoveryDocumentResult.cs
josephdecock: If someone wanted to customize the way we generate discovery documents at a really low level, they could write their own `IHttpResponseWriter<DiscoveryDocumentResult>` and expect the `Entries` to be there. It's a pretty unusual thing to do, but we can keep it around at basically no cost, right?

PR #1831, opened 2025-03-05
Add [ExperimentalAttribute] on preview feature toggles

maartenba:
This PR goes on top of #1830 and adds the [ExperimentalAttribute] to feature preview toggles.

When a preview feature toggle is used in consuming code, a diagnostic is reported and has to be suppressed to explicitly acknowledge using the toggle:

image

To disable, the developer can:
* Add `<PropertyGroup><NoWarn>DUENDEPREVIEW001</NoWarn></PropertyGroup>` to the project file
* Add `#pragma warning disable DUENDEPREVIEW001` at the call site

The diagnostic will also log a link to `https://duende.link/previewfeatures?id={0}` (with the format string replaced by the diagnostic) so we can explain what the feature preview is, how to opt-in, etc.

Background in https://blog.maartenballiauw.be/post/2023/11/08/opt-in-to-experimental-apis-using-csharp-12-experimentalattribute.html.

### DIFF
--- a/identity-server/src/IdentityServer/Configuration/DependencyInjection/Options/PreviewFeaturesOptions.cs
+++ b/identity-server/src/IdentityServer/Configuration/DependencyInjection/Options/PreviewFeaturesOptions.cs
@@ -1,3 +1,5 @@
+using System.Diagnostics.CodeAnalysis;
+
 namespace Duende.IdentityServer.Configuration;
 
 /// <summary>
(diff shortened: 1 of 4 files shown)


identity-server/src/IdentityServer/Configuration/DependencyInjection/Options/PreviewFeaturesOptions.cs
```diff
@@ -8,6 +10,7 @@ public class PreviewFeaturesOptions
     /// <summary>
     /// Enables Caching of Discovery Document based on ResponseCaching Interval 
     /// </summary>
+    [Experimental("DUENDEPREVIEW001", UrlFormat = "https://duende.link/previewfeatures?id={0}")]
```
khalidabuhakmeh, line 13: Who is building this endpoint? 😅

maartenba: Easy!

https://github.com/DuendeSoftware/fenix/pull/5
https://github.com/DuendeSoftware/docs.duendesoftware.com/pull/611

PR #1842, opened 2025-03-06
Enforce final newline in editor config

josephdecock:


bhazen (approved):

josephdecock:
@bhazen do you think there is something that we can do about the codeql complaints?

bhazen:
> @bhazen do you think there is something that we can do about the codeql complaints?

I dismissed the one spot it was flagging yesterday, but the configuration complaint is something we need to fix. Let's look at that together at some point today.

PR #1845, opened 2025-03-07
Use aspire to run IdentityServer development hosts and clients 

StuFrankish:
Added Aspire orchestration using latest version of the Aspire Workload. Refactored IdentityServer Hosts (with migrations), Clients and API projects. Removed some interactivity from console clients as Aspire is not really interactive-console app friendly.

Introducing Aspire orchestration is a natural step to improving the developer experience, streamlining the management of interdependent services and configurations. By automating tasks like service discovery, connection string handling, and dependency orchestration, it reduces manual setup overhead and lets developers focus on core business logic. This unified approach leads to a more consistent, resilient, and observable local development environment.

The Aspire orchestration introduced here has a couple of key call-outs:
- The Aspire Host itself contains additional configuration options in `appsettings.json` to allow for Identity Server Host selection, multiple client selection and multiple API resource selection. Simply select which host, clients and API's you want to run, and launch the Aspire Host.
- Identity Server host projects depending on database functionality now use SQL Server (replacing SqlLite) using Aspire's Docker support, bringing the samples more inline with customer environments.

**Notes:**
While the Web Application clients will startup automatically, the console clients have been amended to run as autonomously as possible and have been set to use explicit start, meaning you will need to manually start them as needed.
These clients will start, execute and shutdown once completed.

As SQL Server has been introduced, you will need a Docker service running on your machine. Aspire will warn if no docker service is running.


josephdecock (changes requested):
I'm still getting build errors out of the aspire.orchestrator project on my machine.

It may be that the build problems are local to me, but we should prove that by building the orchestration project during CI. To do that, we need to add the new projects to the identity-server/identity-server.slnf solution filter file. All of the builds in ci are using the slnf files. 

One other thing I notice is that you chose different names for the orchestration components than Erwin did on the bff side. Do you think we would get any benefit by having them be more similar? If so, should we could change either this or the bff stuff to match.

josephdecock (approved):

josephdecock:
@Erwinvandervalk: I'm going to take the liberty of merging this without your review even though it touches some BFF files, because the changes are really trivial. What we've done is, the package versions are now controlled without conditionally set msbuild properties. With that gone, it is no longer necessary to have our weird use of <TargetFrameworks> even when there is only 1 target framework as a work-around for other tooling issues. I did a mass replace of singleton target frameworks, and a bunch of bff hosts got touched.

Erwinvandervalk:
@josephdecock thanks! all good!

PR #1855, opened 2025-03-10
format proj and template config files

josephdecock:


bhazen (approved):

.editorconfig
```diff
@@ -18,3 +19,5 @@ dotnet_diagnostic.IDE0005.severity = warning
 # Sort system usings first
 dotnet_sort_system_directives_first = true
 
+[*.{json,csproj}]
+indent_size = 2
```
Erwinvandervalk, line 23: Why go for an indent size of 4 for json and csproj for a smaller indent size?

josephdecock: This is what I'm used to, but it is entirely a matter of taste. A lot of the dotnet projects, such as runtime and aspnetcore, use this convention.

Erwinvandervalk: then let's go for it

Erwinvandervalk (approved):

.editorconfig
```diff
@@ -1,15 +1,16 @@
+[*]
+indent_style = space
+indent_size = 4
```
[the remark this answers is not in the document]
josephdecock: I agree, thanks for the suggestion!

PR #1857, opened 2025-03-11
DefaultCache<T> should use a keyed lock instead of a global lock for each T

maartenba:
Debugging what seemed like a race condition resulted in discovering `DefaultCache<T>` locks globally instead of per cache key. Especially in the client stores this is an issue, as retrieving and trying to cache clientA results in requests for clientB having to wait for the lock to become available.

In addition ,we found that some dependencies were registered as transient where a longer lifetime makes more sense.

This PR serves as a discussion:
* Is locking per cache key better than locking the cache altogether?
* Do we want to expand lifetimes where we changed them in this PR?

### DIFF
--- a/identity-server/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/Core.cs
+++ b/identity-server/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/Core.cs
@@ -202,6 +202,7 @@ public static class IdentityServerBuilderExtensionsCore
         builder.Services.TryAddTransient<IBackchannelAuthenticationUserValidator, NopBackchannelAuthenticationUserValidator>();
 
         builder.Services.TryAddTransient(typeof(IConcurrencyLock<>), typeof(DefaultConcurrencyLock<>));
+        builder.Services.TryAddSingleton(typeof(IKeyedConcurrencyLock<>), typeof(DefaultKeyedConcurrencyLock<>));
 
         builder.Services.TryAddTransient<IClientStore, EmptyClientStore>();
         builder.Services.TryAddTransient<IResourceStore, EmptyResourceStore>();
--- a/identity-server/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/InMemory.cs
+++ b/identity-server/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/InMemory.cs
@@ -26,7 +26,7 @@ public static class IdentityServerBuilderExtensionsInMemory
     public static IIdentityServerBuilder AddInMemoryCaching(this IIdentityServerBuilder builder)
     {
         builder.Services.TryAddSingleton<IMemoryCache, MemoryCache>();
-        builder.Services.TryAddTransient(typeof(ICache<>), typeof(DefaultCache<>));
+        builder.Services.TryAddSingleton(typeof(ICache<>), typeof(DefaultCache<>));
 
         return builder;
     }
--- a/identity-server/src/IdentityServer/Infrastructure/ConcurrencyLock/DefaultKeyedConcurrencyLock.cs
+++ b/identity-server/src/IdentityServer/Infrastructure/ConcurrencyLock/DefaultKeyedConcurrencyLock.cs
@@ -1,0 +1,38 @@
+// Copyright (c) Duende Software. All rights reserved.
+// See LICENSE in the project root for license information.
+
+
+using System.Collections.Concurrent;
+
+namespace Duende.IdentityServer.Internal;
+
+/// <summary>
+/// Represents a default implementation of a keyed concurrency lock,
+/// allowing tasks to acquire and release locks based on specific keys.
+/// </summary>
+/// <typeparam name="T">The type associated with the keyed concurrency lock.</typeparam>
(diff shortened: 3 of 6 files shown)


khalidabuhakmeh:
This is another proposed solution to this issue that does not introduce yet another interface that users could accidentally override or change.

## InMemory.cs

```diff
     public static IIdentityServerBuilder AddInMemoryCaching(this IIdentityServerBuilder builder)
     {
         builder.Services.TryAddSingleton<IMemoryCache, MemoryCache>();
-        builder.Services.TryAddTransient(typeof(ICache<>), typeof(DefaultCache<>));
+        builder.Services.TryAddSingleton(typeof(ICache<>), typeof(DefaultCache<>));
 
         return builder;
     }
```

## DefaultCache

```csharp
// See LICENSE in the project root for license information.


using Duende.IdentityServer.Configuration;
using Microsoft.Extensions.Caching.Memory;
using Microsoft.Extensions.Logging;

namespace Duende.IdentityServer.Services;

/// <summary>
/// IMemoryCache-based implementation of the cache
/// </summary>
/// <typeparam name="T"></typeparam>
/// <seealso cref="ICache{T}" />
public class DefaultCache<T> : ICache<T>
    where T : class
{
    private readonly object _lock = new();

    private const string KeySeparator = "-";

    /// <summary>
    /// The IdentityServerOptions.
    /// </summary>
    public IdentityServerOptions IdentityServerOptions { get; }

    /// <summary>
    /// The memory cache.
    /// </summary>
    protected IMemoryCache Cache { get; }

    /// <summary>
    /// The logger.
    /// </summary>
    protected ILogger<DefaultCache<T>> Logger { get; }

    /// <summary>
    /// Initializes a new instance of the <see cref="DefaultCache{T}"/> class.
    /// </summary>
    /// <param name="identityServerOptions"></param>
    /// <param name="cache">The cache.</param>
    /// <param name="logger">The logger.</param>
    public DefaultCache(IdentityServerOptions identityServerOptions, IMemoryCache cache,
        ILogger<DefaultCache<T>> logger)
    {
        IdentityServerOptions = identityServerOptions;
        Cache = cache;
        Logger = logger;
    }

    /// <summary>
    /// Used to create the key for the cache based on the data type being cached.
    /// </summary>
    /// <param name="key"></param>
    /// <returns></returns>
    protected string GetKey(string key)
    {
        return $"{typeof(T).FullName}{KeySeparator}{key}";
    }

    /// <inheritdoc/>
    public Task<T> GetAsync(string key)
    {
        using var activity = Tracing.CacheActivitySource.StartActivity("DefaultCache.Get");

        key = GetKey(key);
        var item = Cache.Get<T>(key);
        return Task.FromResult(item);
    }

    /// <inheritdoc/>
    public Task SetAsync(string key, T item, TimeSpan expiration)
    {
        using var activity = Tracing.CacheActivitySource.StartActivity("DefaultCache.Set");

        key = GetKey(key);
        Cache.Set(key, item, expiration);
        return Task.CompletedTask;
    }

    /// <inheritdoc/>
    public Task RemoveAsync(string key)
    {
        using var activity = Tracing.CacheActivitySource.StartActivity("DefaultCache.Remove");

        key = GetKey(key);
        Cache.Remove(key);
        return Task.CompletedTask;
    }

    /// <inheritdoc/>
    public async Task<T> GetOrAddAsync(string key, TimeSpan duration, Func<Task<T>> get)
    {
        using var activity = Tracing.CacheActivitySource.StartActivity("DefaultCache.GetOrAdd");

        ArgumentNullException.ThrowIfNull(key);
        ArgumentNullException.ThrowIfNull(get);

        var item = await GetAsync(key);

        if (item == null)
        {
            if (Monitor.TryEnter(_lock, (int)IdentityServerOptions.Caching.CacheLockTimeout.TotalMilliseconds))
            {
                try
                {
                    // double check
                    item = await GetAsync(key);

                    if (item == null)
                    {
                        Logger.LogTrace("Cache miss for {cacheKey}", key);

                        item = await get();

                        if (item != null)
                        {
                            Logger.LogTrace("Setting item in cache for {cacheKey}", key);
                            await SetAsync(key, item, duration);
                        }
                    }
                    else
                    {
                        Logger.LogTrace("Cache hit for {cacheKey}", key);
                    }
                }
                finally
                {
                    Monitor.Exit(_lock);
                }
            }
            else
            {
                throw new Exception($"Failed to obtain cache lock for: '{GetType()}'");
            }
        }
        else
        {
            Logger.LogTrace("Cache hit for {cacheKey}", key);
        }

        return item;
    }
}
```

maartenba:
Added test for the new implementation. It succeeds (yay!), and if you run it against the previous implementation it hangs for a while and then crashes with the cache lock exception.

josephdecock (changes requested):
Thanks Maarten! This is an interesting idea, and warrants further investigation. There's some other thoughts going around about experimenting with `HybridCache` as well.

If we take this approach, it would be a breaking change for customizations that derived from `DefaultCache`, so it would be something we do in a major release. Merging now would put it into the 7.3 release, so I'm marking this as "request changes", just in case.

maartenba:
Thanks! Lets revisit during 7.3

josephdecock:
I'm going to close this for now. We should definitely revisit for 8.0.

PR #1859, opened 2025-03-12
Use `DefaultConcurrencyLock` as a Singleton instead of Transient.

maartenba:
Changed the DI registration of `IConcurrencyLock<>` to Singleton to ensure consistent state for `SemaphoreSlim` locks. Also removed the static modifier from `SemaphoreSlim` to align its lifecycle with the class instance.


Erwinvandervalk (approved):

josephdecock (approved):

josephdecock:
Thanks @maartenba! I've reassigned to me, and we will merge this after the editorconfig changes go in, to avoid churn on those.

PR #1860, opened 2025-03-12
Introduce the `StrictClientAssertionAudienceValidation` preview feature

StuFrankish:
This PR introduces the new `StrictClientAssertionAudienceValidation` preview flag.
The preview will enable users to optionally enforce a `typ` header requirement for client authentication jwts while also allowing clients to voluntarily opt into strict validation by setting the `typ` header value to `client-authentication+jwt`.

**Setting the flag within the IdentityServer host configuration**
In a typical IdentityServer host, users may set the preview flag through `options.Preview` feature flags, for example;
```csharp
builder.Services.AddIdentityServer(options =>
{
      options.Preview.StrictClientAssertionAudienceValidation = true;
})
```
Setting this flag will mean that when clients authenticate with private_key_jwt assertions, IdentityServer will validate the audience of the assertion strictly, ie; the audience must be this IdentityServer's issuer identifier as a single string.

### DIFF
--- a/identity-server/src/IdentityServer/Validation/Default/PrivateKeyJwtSecretValidator.cs
+++ b/identity-server/src/IdentityServer/Validation/Default/PrivateKeyJwtSecretValidator.cs
@@ -66,12 +66,12 @@ public class PrivateKeyJwtSecretValidator : ISecretValidator
         // Decide whether to enforce strict audience validation or not.
         var enforceStrictAud = _options.Preview.StrictClientAssertionAudienceValidation;
 
-        if (enforceStrictAud && parsedSecret.Type != "client-authentication+jwt")
+        if (enforceStrictAud && parsedSecret.Type != ParsedSecretTypes.ClientAuthenticationJwt)
         {
             return fail;
         }
 
-        if (parsedSecret.Type != ParsedSecretTypes.JwtBearer && parsedSecret.Type != "client-authentication+jwt")
+        if (parsedSecret.Type != ParsedSecretTypes.JwtBearer && parsedSecret.Type != ParsedSecretTypes.ClientAuthenticationJwt)
         {
             return fail;
         }
@@ -101,7 +101,7 @@ public class PrivateKeyJwtSecretValidator : ISecretValidator
 
         // If strict mode is not enabled by option but the type value "client-authentication+jwt" is provided,
         // enforce strict audience validation.
-        if (string.Equals(parsedSecret.Type, "client-authentication+jwt", StringComparison.OrdinalIgnoreCase))
+        if (string.Equals(parsedSecret.Type, ParsedSecretTypes.ClientAuthenticationJwt, StringComparison.OrdinalIgnoreCase))
         {
             enforceStrictAud = true;
         }
--- a/identity-server/src/IdentityServer/Validation/Default/PrivateKeyJwtSecretValidator.cs
+++ b/identity-server/src/IdentityServer/Validation/Default/PrivateKeyJwtSecretValidator.cs
@@ -145,7 +145,7 @@ public class PrivateKeyJwtSecretValidator : ISecretValidator
                        AudiencesMatch(audString, issuer);
             };
 
-            // Strict audience validation requires that the token type be "JWT"
+            // Strict audience validation requires that the token type be "client-authentication+jwt"
             tokenValidationParameters.ValidTypes = ["client-authentication+jwt"];
         }
         else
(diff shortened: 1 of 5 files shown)


identity-server/src/IdentityServer/Validation/Default/PrivateKeyJwtSecretValidator.cs
```diff
@@ -91,21 +93,44 @@ public async Task<SecretValidationResult> ValidateAsync(IEnumerable<Secret> secr
 
         var issuer = await _issuerNameService.GetCurrentAsync();
 
+        // Decide whether to enforce strict audience validation or not.
+        var enforceStrictAud = _options.Preview.StrictClientAssertionAudienceValidation;
```
khalidabuhakmeh, line 97: This should have a pragma to disable this warning/error or else the build would fail, right?

StuFrankish: Pragma to disable the warning is at the top of the file, happy to bring it closer if that makes things clearer for users.

identity-server/src/IdentityServer/Validation/Default/PrivateKeyJwtSecretValidator.cs
```diff
@@ -1,18 +1,20 @@
 // Copyright (c) Duende Software. All rights reserved.
 // See LICENSE in the project root for license information.
 
+// Supress the warning for the preview feature `Preview.StrictClientAssertionAudienceValidation`
+#pragma warning disable DUENDEPREVIEW002
```
khalidabuhakmeh, line 5: Missed this line. It's happening at the top of the file. 👍🏽

khalidabuhakmeh (approved):

identity-server/src/IdentityServer/Validation/Default/PrivateKeyJwtSecretValidator.cs
```diff
@@ -63,7 +63,15 @@ public async Task<SecretValidationResult> ValidateAsync(IEnumerable<Secret> secr
         var fail = new SecretValidationResult { Success = false };
         var success = new SecretValidationResult { Success = true };
 
-        if (parsedSecret.Type != ParsedSecretTypes.JwtBearer)
+        // Decide whether to enforce strict audience validation or not.
+        var enforceStrictAud = _options.Preview.StrictClientAssertionAudienceValidation;
+
+        if (enforceStrictAud && parsedSecret.Type != "client-authentication+jwt")
```
khalidabuhakmeh, line 69: Should the string `client-authentication+jwt` be put into a constant since it's used in two locations?

identity-server/src/IdentityServer/Validation/Default/PrivateKeyJwtSecretValidator.cs
```diff
@@ -119,9 +136,13 @@ public async Task<SecretValidationResult> ValidateAsync(IEnumerable<Secret> secr
                 return audValue is string audString &&
                        AudiencesMatch(audString, issuer);
             };
+
+            // Strict audience validation requires that the token type be "JWT"
+            tokenValidationParameters.ValidTypes = ["JWT"];
```
josephdecock, line 141: This is the old type value. With strict validation, we want the new value here: "client-authentication+jwt".

identity-server/src/IdentityServer/Validation/Default/PrivateKeyJwtSecretValidator.cs
```diff
@@ -61,7 +63,15 @@ public async Task<SecretValidationResult> ValidateAsync(IEnumerable<Secret> secr
         var fail = new SecretValidationResult { Success = false };
         var success = new SecretValidationResult { Success = true };
 
-        if (parsedSecret.Type != ParsedSecretTypes.JwtBearer)
+        // Decide whether to enforce strict audience validation or not.
+        var enforceStrictAud = _options.Preview.StrictClientAssertionAudienceValidation;
+
+        if (enforceStrictAud && parsedSecret.Type != ParsedSecretTypes.ClientAuthenticationJwt)
```
josephdecock, line 69: We can't use the `parsedSecret.Type` field in this way. That type is a property that is saying what kind of authentication the client has used. If it happens to be a jwt, then we still need to validate the jwt to get at its typ header.

josephdecock (approved):
Looks good! I went ahead and updated some of the test cases I had written for part 1 of the whole 7523bis work as well.

PR #1861, opened 2025-03-12
transform logout url in transformer

Erwinvandervalk:


bff/hosts/Blazor/PerComponent/Hosts.Bff.Blazor.PerComponent/Program.cs
```diff
@@ -42,6 +42,10 @@
     {
         options.Cookie.Name = "__Host-blazor";
         options.Cookie.SameSite = SameSiteMode.Strict;
+        options.Events.OnValidatePrincipal += context =>
```
josephdecock, line 45: Maybe this empty event isn't needed at all?

bff/src/Bff.Blazor/BffBuilderExtensions.cs
```diff
@@ -14,7 +14,7 @@ public static BffBuilder AddBlazorServer(this BffBuilder builder)
         builder.Services
             .AddOpenIdConnectAccessTokenManagement()
             .AddBlazorServerAccessTokenManagement<ServerSideTokenStore>()
-            .AddScoped<AuthenticationStateProvider, BffServerAuthenticationStateProvider>();
+            .AddTransient<IClaimsTransformation, AddServerManagementClaimsTransform>();
```
josephdecock, line 17: Does the transformation need to be created on demanding this, or could it be singleton? I guess this way you get an up to date options instance in the transformation, but maybe that could use an options monitor to get the same behavior.

bff/hosts/Blazor/PerComponent/Hosts.Bff.Blazor.PerComponent/Program.cs
Erwinvandervalk: correct.

bff/src/Bff.Blazor/BffBuilderExtensions.cs
Erwinvandervalk: Also works.

josephdecock (approved):

PR #1862, opened 2025-03-12
Validate that "RememberMe" is allowed.

StuFrankish:
Validate that the "Remember Me" option is allowed before passing the flag into the SignIn Manager to prevent "Remember Me" being set when it's disallowed.


khalidabuhakmeh (approved):

josephdecock (commented):
Good catch! Do we need a similar fix in the non-aspnetidentity hosts? Also, can you please update the templates as well?

StuFrankish:
> Good catch! Do we need a similar fix in the non-aspnetidentity hosts? Also, can you please update the templates as well?

I've updated the aspnet-identity template now too. Only the aspnet-identity host & template needed a fix.
Our EF & InMem hosts and templates all already validate the field correctly using;
```csharp
// only set explicit expiration here if user chooses "remember me". 
// otherwise we rely upon expiration configured in cookie middleware.
var props = new AuthenticationProperties();
if (LoginOptions.AllowRememberLogin && Input.RememberLogin)
{
    props.IsPersistent = true;
    props.ExpiresUtc = DateTimeOffset.UtcNow.Add(LoginOptions.RememberMeLoginDuration);
}
```

josephdecock (approved):

PR #1876, opened 2025-03-15
EditorConfig: Enforce that 'this.' should not be used when it's redundant

damianh:
**What issue does this PR address?**

Enforce that 'this.' should not be used when it's redundant.

This PR is stacked on #1869 and will be rebased when that is merged.

**Important: Any code or remarks in your Pull Request are under the following terms:**

If You provide us with any comments, bug reports, feedback, enhancements, or modifications proposed or suggested by You for the Software, such Feedback is provided on a non-confidential basis (notwithstanding any notice to the contrary You may include in any accompanying communication), and Licensor shall have the right to use such Feedback at its discretion, including, but not limited to the incorporation of such suggested changes into the Software. You hereby grant Licensor a perpetual, irrevocable, transferable, sublicensable, nonexclusive license under all rights necessary to incorporate and use your Feedback for any purpose, including to make and sell any products and services.

(see [our license](https://duendesoftware.com/license/identityserver.pdf), section 7)

PR #1877, opened 2025-03-15
EditorConfig: Enforce using built-in type keywords (e.g., 'string' instead of 'String')

damianh:

- Enforce using built-in type keywords (e.g., 'string' instead of 'String')
- Applies fixes for rule

Stacked on #1876 will be rebased.

### DIFF
--- a/identity-server/src/Storage/Models/Client.cs
+++ b/identity-server/src/Storage/Models/Client.cs
@@ -1,7 +1,6 @@
 // Copyright (c) Duende Software. All rights reserved.
 // See LICENSE in the project root for license information.
 
-
 #nullable enable
 
 using System.Collections;
--- a/.editorconfig
+++ b/.editorconfig
@@ -19,6 +19,11 @@ dotnet_style_qualification_for_property = false:error
 # Remove unnecessary using directives
 dotnet_diagnostic.IDE0005.severity = warning
 
+# Use language keywords instead of framework type names for type references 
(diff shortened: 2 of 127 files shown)


josephdecock (approved):

adamralph (commented):
The change to `.editorconfig` seems to have disappeared.

damianh:
Rebasing atm

damianh:
Added
`dotnet_style_predefined_type_for_locals_parameters_members = true`
`dotnet_style_predefined_type_for_member_access = true`

identity-server/src/Storage/Models/Client.cs
```diff
@@ -452,7 +451,11 @@ private ICollection<string> Clone()
         private ICollection<string> CloneWith(params string[] values)
         {
             var clone = Clone();
-            foreach (var item in values) clone.Add(item);
+            foreach (var item in values)
+            {
+                clone.Add(item);
+            }
+
```
adamralph, line 458: Unrelated, but I'm OK with including it. I wonder why this didn't change in the previous PRs. 🤔

adamralph (approved):

identity-server/src/Storage/Models/Client.cs
damianh: I'm running resharper tool to apply changes. It'll occasionally it'll reformat other things in a file so not intentional. I'll revert this to keep the pr clean.

PR #1878, opened 2025-03-15
EditorConfig: Enforce BSD/Allman braces.

damianh:
**What issue does this PR address?**

Enforce BSD/Allman braces.

- Places braces on new line for all constructs
- Require braces for all if statement
- Keep empty braces {} on same line
- Don't allow blank line after open brace
- Formatting fixes as a result.

Stacked on #1877 will be rebased accordingly.

**Important: Any code or remarks in your Pull Request are under the following terms:**

If You provide us with any comments, bug reports, feedback, enhancements, or modifications proposed or suggested by You for the Software, such Feedback is provided on a non-confidential basis (notwithstanding any notice to the contrary You may include in any accompanying communication), and Licensor shall have the right to use such Feedback at its discretion, including, but not limited to the incorporation of such suggested changes into the Software. You hereby grant Licensor a perpetual, irrevocable, transferable, sublicensable, nonexclusive license under all rights necessary to incorporate and use your Feedback for any purpose, including to make and sell any products and services.

(see [our license](https://duendesoftware.com/license/identityserver.pdf), section 7)

PR #1879, opened 2025-03-15
EditorConfig: use var by default

damianh:

- Enforce 'var' usage for all local variables
- Applies rule fixes

Stacked on #1878 and will be rebased accordingly

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -16,6 +16,12 @@ dotnet_style_qualification_for_field = false:error
 dotnet_style_qualification_for_method = false:error
 dotnet_style_qualification_for_property = false:error
 
+# 'var' preference
+dotnet_diagnostic.IDE0007.severity = error
+csharp_style_var_for_built_in_types = true
+csharp_style_var_when_type_is_apparent = true
+csharp_style_var_elsewhere = true
+
 # Remove unnecessary using directives
 dotnet_diagnostic.IDE0005.severity = warning
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -16,11 +16,11 @@ dotnet_style_qualification_for_field = false:error
 dotnet_style_qualification_for_method = false:error
 dotnet_style_qualification_for_property = false:error
 
-# 'var' preference
+# Use var instead of explicit type
 dotnet_diagnostic.IDE0007.severity = error
+csharp_style_var_elsewhere = true
 csharp_style_var_for_built_in_types = true
 csharp_style_var_when_type_is_apparent = true
-csharp_style_var_elsewhere = true
 
 # Remove unnecessary using directives
 dotnet_diagnostic.IDE0005.severity = warning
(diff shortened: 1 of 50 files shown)


.editorconfig
```diff
@@ -17,3 +17,20 @@ dotnet_sort_system_directives_first = true
...
+
+csharp_new_line_before_open_brace = all # BSD/Allman braces: Place braces on a new line for all constructs
+dotnet_diagnostic.IDE0011.severity = error # Require braces for all if statements
+dotnet_diagnostic.IDE0055.severity = error # Keep empty braces { } on the same line
+csharp_style_allow_blank_line_after_open_brace = false
+csharp_style_allow_blank_line_between_consecutive_braces = false
+
+csharp_style_var_elsewhere = true:error # Enforce 'var' usage for all local variables
+csharp_style_var_when_type_is_apparent = true:error
+csharp_style_var_for_built_in_types = true:error
```
adamralph, line 36: These are options for IDE0007, so they should appear below a setting for that rule.

adamralph (changes requested):

josephdecock (approved):

josephdecock:
(Approving to indicate consensus on the rule, not disagree with Adam's formatting suggestions of the editconfig)

identity-server/src/IdentityServer/ResponseHandling/Default/DiscoveryResponseGenerator.cs
```diff
@@ -393,7 +393,7 @@ where scope.ShowInDiscoveryDocument
         // custom entries
         if (!IEnumerableExtensions.IsNullOrEmpty(Options.Discovery.CustomEntries))
         {
-            foreach ((string key, object value) in Options.Discovery.CustomEntries)
+            foreach ((var key, var value) in Options.Discovery.CustomEntries)
```
adamralph, line 396:
```suggestion
            foreach (var (key, value) in Options.Discovery.CustomEntries)
```

adamralph (changes requested):

.editorconfig
```diff
@@ -16,6 +16,12 @@ dotnet_style_qualification_for_field = false:error
 dotnet_style_qualification_for_method = false:error
 dotnet_style_qualification_for_property = false:error
 
+# 'var' preference
+dotnet_diagnostic.IDE0007.severity = error
+csharp_style_var_for_built_in_types = true
+csharp_style_var_when_type_is_apparent = true
+csharp_style_var_elsewhere = true
```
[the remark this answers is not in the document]
adamralph: I suppose it's a little odd to put `elsewhere` at the top. If it bothers us enough, we _could_ consider listing the options in the order they are documented instead. But if we do choose to do that, we should do it across the whole file, and not just in this section.

Erwinvandervalk (approved):

.editorconfig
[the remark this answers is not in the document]
damianh: Let's go with alphabetical. (the original order is as-documented)

identity-server/src/IdentityServer/ResponseHandling/Default/DiscoveryResponseGenerator.cs
damianh: Oh that's nice. R# didn't pick up on that.

PR #1897, opened 2025-03-20
enable nullable in AuthorizeEndpointBase

SimonCropp:


Erwinvandervalk:

As of the nullability changes that we're going to apply are going to mean breaking api changes, so we have to time them to align with a major version release.

PR #1899, opened 2025-03-20
enable nullable in AuthorizeCallbackEndpoint

SimonCropp:


Erwinvandervalk:

As of the nullability changes that we're going to apply are going to mean breaking api changes, so we have to time them to align with a major version release.

SimonCropp:
can u give me an example of how this is a breaking change?  

usually adding nullability means people get the wanring at compile time instead of breaking at runtime

Erwinvandervalk:
This particular PR is not a breaking change. However, in some places the changes we want to make with regards to nullability are going to mean breaking changes. We've decided to look into enforcing nullable changes as a team when we think the time is right for this and not to perform or accept changes with regards nullability in individual places in the codebase.

SimonCropp:
"we plan on making breaking changes regarding nullability. so we wont accept any nullability related PRs that are not breaking changes"

seems like a strange stance to take. but ok

PR #1900, opened 2025-03-20
remove redundant loggers

SimonCropp:


josephdecock:
I've removed the changes to the BFF and rebased. The BFF changes were technically breaking, because they changed a public constructor's arguments.

PR #1902, opened 2025-03-20
remove redundant null checks from select result

SimonCropp:


josephdecock:
Thanks for the contribution, but I'm closing this PR. It looks like these changes have merge conflicts with your other PR that removed redundant `ToArray` calls. I also really don't want "." as a commit message. Of course we could solve merge conflicts and rebase away the ".", but this seems like too minor of a change to go through that hassle.

SimonCropp:
the . commit message is irrelevant because u should always be doing a squash when u merge PRs in github. then u get on nice commit message the same as the title of the pr

PR #1903, opened 2025-03-20
use null! instead of default!

SimonCropp:


Erwinvandervalk:

In this particular case, it's a style question. We're currently working through standardizing our code style using editor config changes.

PR #1915, opened 2025-03-21
use some deconstruction

SimonCropp:


josephdecock:
Thanks, but this is a purely code-style change and we're not currently accepting community PRs that only make style changes.

PR #1916, opened 2025-03-21
remove redundant return values

SimonCropp:


josephdecock:
Even though these aren't being used right now, I'd prefer to keep them, for consistency and to make it trivial to compose these setup methods in future.

PR #1920, opened 2025-03-24
Adds a dotnet 9 test host

StuFrankish:
Adds a dedicated dotnet 9 test host based on the existing Entity Framework host. This allows us to run our dev environment entirely against dotnet 9 while still supporting dotnet 8.


identity-server/clients/src/ConsoleDcrClient/Program.cs
```diff
@@ -20,13 +20,21 @@
 })
 .AddServiceDiscovery();
 
+var clientId = Guid.NewGuid().ToString();
```
bhazen, line 23: Was this change intended to be part of this PR?

identity-server/clients/src/ConsoleClientCredentialsFlowDPoP/Program.cs
```diff
@@ -14,8 +14,10 @@
 // Add ServiceDefaults from Aspire
 builder.AddServiceDefaults();
 
-var discoClient = new HttpClient();
 var authority = builder.Configuration["is-host"];
+var simpleApi = builder.Configuration["simple-api"];
```
josephdecock, line 18: Let's remove this, since simple-api isn't used.

PR #1926, opened 2025-03-26
Don't unnecessarily return scope parameter from authorize endpoint

bhazen:
This addresses a customer reported bug which was transferred to the internal backlog. The customer reported:
"We’ve noticed that the scope parameter is included in the callback URL after authentication in the authorization code flow, also when using PAR. This behavior seems unexpected, as scope is already handled during the authorization process and doesn’t appear to be required in the callback URL. In particular with PAR, where one of the benefits should be shorter URLs.

Wondering if this is an intentional design choice? If so, what purpose does it serve?"

Per RFC 6749, the only time the authorize endpoint needs to return the scope parameter is during the implicit flow as defined in [section 4.2.2](https://datatracker.ietf.org/doc/html/rfc6749#section-4.2.2). This change prevents the scope parameter from being included in a response from the authorize endpoint unless a token is issued as it the scope parameter will be returned from the token endpoint for other grant types.


josephdecock (approved):
Nice work fixing a bug by moving curly braces 😄

PR #1927, opened 2025-03-28
Allow all clients to run against both EF hosts (net8.0 and net9.0)

StuFrankish:
Allows all clients to operate against the EF (net8.0 & net9.0) hosts.

- Condense duplicated extensions into common hosts extensions link folder.
- Condense duplicated configuration (clients, resources, users) into common hosts configuration link folder.
- Renamed Clients.cs & Resources.cs to include "Test" prefix to avoid conflict with namespaces in the EF hosts.
- Configuration combined from Host_Main & Host_Configuration into both EF hosts to support clients.


identity-server/hosts/EntityFramework-dotnet9/TestKeys/identityserver.test.ecdsa.p12
josephdecock, line 1: I would expect that neither of these keys would be usable outside of your system, because when you generated them with mkcert, mkcert signed them with your local mkcert signing key. Even if other devs import these keys into their keychains, they can't (well shouldn't) import your mkcert signing key as a trusted root signing authority. 

I think the existing places where we have static keys checked in are from ancient days when there was just Dom working with them, and he checked in the keys for his own convenience. Today, everyone needs to gen their own key, so let's remove these 2 key files. If you think more instructions would help, we could add that, but there are some existing comments saying how to use mkcert.

identity-server/hosts/EntityFramework/Host.EntityFramework.csproj
```diff
@@ -34,13 +34,53 @@
...
-    <ProjectReference
-      Include="..\..\src\Configuration.EntityFramework\Duende.IdentityServer.Configuration.EntityFramework.csproj" />
-    <ProjectReference
-      Include="..\..\src\EntityFramework\Duende.IdentityServer.EntityFramework.csproj" />
+    <ProjectReference Include="..\..\src\Configuration.EntityFramework\Duende.IdentityServer.Configuration.EntityFramework.csproj" />
+    <ProjectReference Include="..\..\src\EntityFramework\Duende.IdentityServer.EntityFramework.csproj" />
+    <ProjectReference Include="..\..\src\IdentityServer\Duende.IdentityServer.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
```
josephdecock, line 44: Another wildcard please.

identity-server/hosts/EntityFramework/IdentityServerExtensions.cs
```diff
@@ -45,14 +70,77 @@ internal static WebApplicationBuilder ConfigureIdentityServer(this WebApplicatio
...
+
+        return builder;
+    }
+
+    // To use static signing credentials, create keys and add it to the certificate store.
+    // This shows how to create both rsa and ec keys, in case you had clients that were configured to use different algorithms
+    // You can create keys for dev use with the mkcert util:
+    //    mkcert -pkcs12 identityserver.test.rsa
+    //    mkcert -pkcs12 -ecdsa identityserver.test.ecdsa
+    // Then import the keys into the certificate manager. This code expect keys in the personal store of the current user.
```
josephdecock, line 116:
```suggestion
    // Then import the keys into the certificate manager. This code expects keys in the personal store of the current user.
```

identity-server/hosts/EntityFramework/TestKeys/identityserver.test.ecdsa.p12
josephdecock, line 1: Also remove these keys.

identity-server/hosts/main/Host.Main.csproj
```diff
@@ -42,17 +42,41 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Compile Include="..\Config\Clients.cs">
-      <Link>Configuration\Clients.cs</Link>
+    <Compile Include="..\Config\TestClients.cs">
```
josephdecock, line 45: Yet another wildcard spot.

identity-server/migrations/IdentityServerDb/IdentityServerDb.csproj
```diff
@@ -31,17 +31,20 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Compile Include="..\..\hosts\Config\Clients.cs">
-      <Link>Configuration\Clients.cs</Link>
+    <Compile Include="..\..\hosts\Config\TestClients.cs">
```
josephdecock, line 34: Might as well do the wildcard thing here too, at least for the Configuration subdir.

identity-server/hosts/EntityFramework/IdentityServerExtensions.cs
```diff
@@ -45,14 +70,77 @@ internal static WebApplicationBuilder ConfigureIdentityServer(this WebApplicatio
             .AddScopeParser<ParameterizedScopeParser>()
             // this is something you will want in production to reduce load on and requests to the DB
             //.AddConfigurationStoreCache()
-            ;
+
+            // IMPORTED EXTENSIONS
```
josephdecock, line 74:
```suggestion
```

josephdecock (approved):

PR #1937, opened 2025-04-02
Implement RFC 9701 to Return JWT Response from Introspection Endpoint

bhazen:
Updates the token introspection endpoint to support the behavior for returning the response as a JWT as defined in [RFC 9701](https://www.rfc-editor.org/rfc/rfc9701.html)


josephdecock (commented):
Looks great! My only concern is that we should verify that an access token's claims might be complex or nested values. Can we add a test to prove that that works?

bhazen:
> Looks great! My only concern is that we should verify that an access token's claims get serialized correctly if they are complex or nested values. Can we add a test to prove that that arrays and nested objects work?

@josephdecock I added a couple more tests at the end of the introspection endpoint tests file. Let me know if that is not what you were looking for with this comment.

PR #1947, opened 2025-04-04
V2 License Functionality Updates

bhazen:
Added setting client and issuer limits to license and added an abstraction of controlling how logging happens to keep the logic centralized.


identity-server/src/IdentityServer/Licensing/V2/LicenseExpirationChecker.cs
```diff
@@ -12,13 +12,13 @@ internal class LicenseExpirationChecker(
...
 {
-    private readonly ILogger _logger = loggerFactory.CreateLogger("Duende.IdentityServer.License");
+    private readonly LicenseLoggingAdapter _logger = new(license.Current.Redistribution, loggerFactory.CreateLogger("Duende.IdentityServer.License"));
 
     private bool _expiredLicenseWarned;
 
     public void CheckExpiration()
     {
-        if (!_expiredLicenseWarned && !license.Current.Redistribution && IsExpired)
+        if (!_expiredLicenseWarned && IsExpired)
```
josephdecock, line 21: I think we have to keep the old language for redist licenses, since we aren't planning license enforcement changes for redist.

bhazen: Ok, it sounds like I misinterpreted part of our discussion about feature parity with the V1 code then. I had thought we were going to redirect all log output for redistribution licenses to Trace as we do with the V1 license code. I added an abstraction to centralize that logic so we didn't need to have it repeated it the business logic of license checks. It sounds like that's not the case and the V2 code should not be logging _at all_ for redistribution licenses. Is that correct? If so, I think it's worth keeping the logging abstraction I put in place and updating it to not emit any log entries of the license is a redistribution license. What are your thoughts on that?

josephdecock: Yes, sorry for the confusion.

The v2 license messages are only for non-redist. We do also want to keep logging the old messages at the trace level. 

In the long term, I'd like to have the v1 code entirely be replaced by v2. So eventually (maybe now?) the redist trace level message will need to be logged by the v2 code. I think the adapter you have in the PR right now is a good idea, but if we add this wrinkle that the messages will be different too, can it still be helpful?

Another option to consider might be the way that Erwin had been evolving the logging in bff and atm.

josephdecock (approved):

PR #1949, opened 2025-04-07
Reorganizing BFF to improve future maintainability

Erwinvandervalk:

Reorganizes the codebase to improve maintainability. 

* [x] Reorganize namespaces
* [x] Mark implementation types as internal
* [x] Rename IBffApiEndpoint to IBffApiMetadata

## No more types in Microsoft namespaces. 

The default extension methods are no longer in the namespace of the system they are enriching. IE: BffApplicationBuilderExtensions was moved from Microsoft.AspNetCore.Builder to Duende.Bff.

This means users need to add an additional using statements to the system. This may reduce discoverability of our functionality. However, this explicit step has two benefits:
* When somebody adds the Bff to a product, it's clear from the namespaces that this has happened. So reviewing code is easier. 
* Putting extension methods in namespaces you own can cause collisions that cause compile issues for our consumers. 

## Marking implementation classes as internal

Our default implementations should not be extensibility points to be derived from. This severely limits our ability to make improvements to the software without doing breaking changes. 


bff/src/Bff/Bff.csproj.DotSettings
```diff
@@ -0,0 +1,3 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
```
josephdecock, line 1: I think we should exclude DotSettings in .gitignore.

bff/src/Bff/Internal/DefaultAccessTokenRetriever.cs
```diff
@@ -1,13 +1,15 @@
...
+namespace Duende.Bff.Internal;
 
 /// <summary>
 /// Default implementation of IAccessTokenRetriever
 /// </summary>
-public class DefaultAccessTokenRetriever() : IAccessTokenRetriever
+internal class DefaultAccessTokenRetriever() : IAccessTokenRetriever
 {
     /// <inheritdoc />
     public virtual async Task<AccessTokenResult> GetAccessToken(AccessTokenRetrievalContext context)
```
josephdecock, line 15: I don't think this needs to be virtual anymore

bff/src/Bff/Endpoints/DefaultUserService.cs
```diff
@@ -3,22 +3,21 @@
...
 
-// ReSharper disable once CheckNamespace
-namespace Duende.Bff;
+namespace Duende.Bff.Endpoints;
 
 /// <summary>
 /// Service for handling user requests
 /// </summary>
-public class DefaultUserService(IOptions<BffOptions> options, ILoggerFactory loggerFactory) : IUserService
+internal class DefaultUserService(IOptions<BffOptions> options, ILoggerFactory loggerFactory) : IUserService
```
josephdecock, line 20: When we make this internal, we should add an extensibility point for customizing how management and user claims are retrieved. (Okay to do as a follow up.)

bff/src/Bff/Endpoints/DefaultBackchannelLogoutService.cs
```diff
@@ -13,13 +15,12 @@
...
 
-// ReSharper disable once CheckNamespace
-namespace Duende.Bff;
+namespace Duende.Bff.Endpoints;
 
 /// <summary>
 /// Default back-channel logout notification service implementation
 /// </summary>
-public class DefaultBackchannelLogoutService(
+internal class DefaultBackchannelLogoutService(
```
josephdecock, line 23: Validation of logout tokens is an extensibility point today, so we should consider adding a replacement for that. Users might want custom token validation parameters, or might want to customize even further to allow logout tokens to be JWEs.

bff/src/Bff/Licensing/LicenseValidatorLocal.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Duende Software. All rights reserved.
```
josephdecock, line 1: Consider renaming this file to match its class name (or is this intentional, because it is a partial?)

josephdecock (approved):

bff/src/Bff/Bff.csproj.DotSettings
Erwinvandervalk: I'll bring it up as a discussion point.

bff/src/Bff/Endpoints/DefaultBackchannelLogoutService.cs
Erwinvandervalk: created issue for this

bff/src/Bff/Internal/DefaultAccessTokenRetriever.cs
Erwinvandervalk: true

PR #1956, opened 2025-04-09
Use PAR IdentityModel constants

josephdecock:
This is a minor cleanup of an old todo.


bhazen (approved):

josephdecock:
> The builds are very confused. I made this commit from a feature branch and then rebased it back to main. Recreating...

Woops, target branch was wrong. Reopening this and fixing.

PR #1974, opened 2025-04-24
Support binding of authorization codes to dpop proof key via dpop header

josephdecock:
This PR adds support for the DPoP header at the PAR endpoint.

The DPoP header at the PAR endpoint is used to bind authorization codes to the client's dpop key. This can also be done by (and we already support) passing the dpop_jkt parameter.

This change improves our support for RFC 9449, as accepting the DPoP header is a requirement for compliance now that we implement PAR (section 10.1). We also now enforce the rule that if both the header and the dpop_jkt parameter are passed, they must be consistent. 

This PR also significantly refactors the existing DPOP integration tests to make them easier to read and reuse. Most behaviors are the same whether authorize parameters are pushed or not, and either or both of the dpop binding mechanisms can be used with PAR, so we want to run through most of those tests in several ways.

This PR updates our error responses to consistently return "use_dpop_nonce" as the error when a request fails and needs a new nonce, even if an invalid or expired nonce was passed.

### DIFF
--- a/identity-server/src/IdentityServer/Validation/Default/PushedAuthorizationRequestValidator.cs
+++ b/identity-server/src/IdentityServer/Validation/Default/PushedAuthorizationRequestValidator.cs
@@ -5,8 +5,12 @@
 #nullable enable
 
 using Duende.IdentityModel;
+using Duende.IdentityServer.Configuration;
 using Duende.IdentityServer.Extensions;
 using Duende.IdentityServer.Licensing.V2;
+using Duende.IdentityServer.Services;
+using Microsoft.Extensions.Logging;
+using static Duende.IdentityServer.IdentityServerConstants;
 
 namespace Duende.IdentityServer.Validation;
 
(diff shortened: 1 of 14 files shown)


identity-server/src/IdentityServer/Validation/Default/PushedAuthorizationRequestValidator.cs
```diff
@@ -16,38 +20,116 @@ namespace Duende.IdentityServer.Validation;
...
+
+            if (dpopValidationResult.Error != null)
+            {
+                return new PushedAuthorizationValidationResult(
+                    dpopValidationResult.Error,
+                    dpopValidationResult.ErrorDescription ?? "Invalid DPoP Proof");
+            }
+
+            // if dpop_jkt was also passed, make sure they are consistent
+            var dpopThumbprintParameter = context.RequestParameters.Get(OidcConstants.AuthorizeRequest.DPoPKeyThumbprint);
```
bhazen, line 111: To make sure I'm understanding part of the problem: we only need to do this if there is a proof token because this is another way of passing the key thumbprint, correct? There's only one way to pass the proof token, but multiple ways to pass the thumbprint? Is there anything we need to do if we only receive the thumbprint passed this way and no proof token?

identity-server/test/IdentityServer.IntegrationTests/Endpoints/Token/DPoPPushedAuthorizationEndpointTests.cs
```diff
@@ -0,0 +1,110 @@
...
+using IntegrationTests.Endpoints.Token;
+
+namespace IntegrationTests.Endpoints.PushedAuthorization;
+
+/// <summary>
+/// This class contains tests of the DPoP support at the PAR endpoint, exercising behavior that is specific to PAR.
+/// Many more tests in DPoPTokenEndpointTests exercise the code flow and are parameterized to test both with and without
+/// pushed authorization.
+/// </summary>
+public class DPoPPushedAuthorizationEndpointTests : DPoPEndpointTestBase
```
bhazen, line 17: It's probably worth adding a test which exercises the successful path here to server as a regression test.

identity-server/src/IdentityServer/Validation/Default/PushedAuthorizationRequestValidator.cs
josephdecock: It's slightly different. If we're binding the auth code to a dpop key, we only really need the thumbprint in order to do that binding. You can pass the thumbprint directly as the dpop_jkt parameter. You can also pass a full dpop proof, which contains the public key - not just its thumbprint - as a jwk. 

If you pass the proof, we derive the thumbprint of the public key within it, and then act as though that value was passed to us as the dpop_jkt (the thumbprint). So, in that way, passing a proof is sort of a second way of passing a thumbprint - at least it is something that a thumbprint can be derived from.

If the two don't agree, that's an error: the client is asking to bind the auth code to 2 different proof keys, which we can't do. 

Finally, yes, there is only one way to pass the dpop proof (via the dpop http header), but the proof only matters to us because it conveys the public key material that we're going to bind the auth code to. If we don't receive the proof but do receive the dpop_jkt parameter, we can just use that parameter.

identity-server/test/IdentityServer.IntegrationTests/Endpoints/Token/DPoPPushedAuthorizationEndpointTests.cs
josephdecock: What about the `DPoPTokenEndpointTests` - `valid_dpop_request_should_return_bound_refresh_token` and `public_client_dpop_proof_should_be_required_on_renewal`? Those get parameterized to do PAR or not, and bind auth codes or not.

bhazen: Ok. I missed those initially with the GitHub web UI hiding that file due to its size.

bhazen (approved):

PR #1980, opened 2025-04-26
Remove global.json

josephdecock:
We no longer need to pin the sdk version, because the fix for the dotnet format issue that originally prompted that has been released.


josephdecock:
We're going to keep pinning the version, but relax the rules for local dev and advance the version in  Closing.

PR #1985, opened 2025-04-27
remove un-used RemoveLeadingSlash

SimonCropp:


khalidabuhakmeh:
@bhazen @josephdecock This Pull Request is no longer valid because it's used now. I think we can close this pull request.

https://github.com/DuendeSoftware/products/blob/9c19a141699ab092f9a24c4be0d320eb8f68e077/identity-server/src/IdentityServer/Services/Default/DefaultServerUrls.cs#L44

josephdecock:
Instead of removing this, I opted to unify the multiple copies of the string extensions. Thanks anyway though!

PR #1988, opened 2025-04-29
Use ServiceDefaults for IdSrv host to get Otel in Aspire

AndersAbel:
Now that we have Aspire for the dev environment I think it makes sense to have OTel completely wired up for those hosts too. I simply called AddServiceDefaults from the IdSrv host. Is that risky? Could there be other side effects?

For now I've just done one host and created this as a draft. If everything is ok I'll make the same changes to other hosts.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -93,7 +93,6 @@
     <PackageVersion Include="NSubstitute" Version="5.1.0" />
     <PackageVersion Include="OpenTelemetry.Exporter.Console" Version="1.11.0" />
     <PackageVersion Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="1.11.0" />
-    <PackageVersion Include="OpenTelemetry.Exporter.Prometheus.AspNetCore" Version="1.11.0-beta.1" />
     <PackageVersion Include="OpenTelemetry.Extensions.Hosting" Version="1.11.0" />
     <PackageVersion Include="OpenTelemetry.Instrumentation.AspNetCore" Version="1.11.0" />
     <PackageVersion Include="OpenTelemetry.Instrumentation.Http" Version="1.11.0" />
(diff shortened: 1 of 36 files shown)


AndersAbel:
@bhazen @josephdecock Would you mind having a look at this? If it looks ok I'll apply corresponding changes to all host and submit the full PR.

bhazen:
> @bhazen @josephdecock Would you mind having a look at this? If it looks ok I'll apply corresponding changes to all host and submit the full PR.

@AndersAbel this looks fine to me. Please proceed with the rest of the changes for the full PR.

josephdecock:
I experimented with this a bit, and I think we may also need to change how we're capturing logs so that the log messages show up in the aspire structured logs. Some services show up there (mvc.code, simple-api) but others do not (is-host). The difference I see between the two is that the is-host uses serilog's two stage log initialization (the call CreateBootstrapLogger before doing anything else, and then later call UseSerilog), while the others serivces don't bother with the bootstrap. I did a little bit of research and saw this: https://github.com/serilog/serilog-aspnetcore/issues/359, so maybe we need to explicitly wire up a serilog WriteToOpenTelemetry? It's a bit unclear to me why structured logging doesn't work with the two stage logger. Maybe aspire is automagically wiring things up, but then the second stage logger comes after aspire does its thing? Anyway, would be great to get both spans and structured logs working in the dashboard.

josephdecock:
Summary of changes

- Add Duende meters in aspire service defaults
- Invoke service defaults in all hosts and apis
- Clean up all hosts and apis 
- Add shared code to set up Serilog bootstrap logger and main logger
- Remove duplicated appsettings in favor of that single shared serilog code
- Removed prometheus and honeycomb in favor of aspire tool

AndersAbel:
> sorry to just take over

No worries, I'm just glad you picked it up and got it completed. I've been busy with other stuff.

PR #1994, opened 2025-05-03
DPoP Jwt Bearer extensibility reworked

josephdecock:
This is a fairly significant refactor of the DPoP jwt bearer extension's extensibility model that makes most of the types internal. The intention is to make the intended extensibility points explicit, rather than just making everything public.

### DIFF
--- a/aspnetcore-authentication-jwtbearer/src/AspNetCore.Authentication.JwtBearer/DPoP/DPoPProofValidator.cs
+++ b/aspnetcore-authentication-jwtbearer/src/AspNetCore.Authentication.JwtBearer/DPoP/DPoPProofValidator.cs
@@ -158,7 +158,7 @@ internal class DPoPProofValidator : IDPoPProofValidator
         }
     }
 
-    internal bool ValidateJwk(DPoPProofValidationContext context, DPoPProofValidationResult result)
+    internal void ValidateJwk(DPoPProofValidationContext context, DPoPProofValidationResult result)
     {
         JsonWebToken token;
 
@@ -171,14 +171,14 @@ internal class DPoPProofValidator : IDPoPProofValidator
         {
             Logger.LogDebug("Error parsing DPoP proof token: {error}", ex.Message);
             result.SetError("Malformed DPoP proof token.");
-            return false;
+            return;
         }
 
         if (!token.TryGetHeaderValue<JsonElement>(JwtClaimTypes.JsonWebKey, out var jwkValues))
         {
             Logger.LogDebug("Failed to get jwk header");
             result.SetError("Invalid 'jwk' value.");
-            return false;
+            return;
         }
 
         var jwkJson = JsonSerializer.Serialize(jwkValues);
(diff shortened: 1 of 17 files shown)


aspnetcore-authentication-jwtbearer/src/AspNetCore.Authentication.JwtBearer/DPoP/DPoPProofValidator.cs
```diff
@@ -89,76 +70,115 @@ public async Task<DPoPProofValidationResult> Validate(DPoPProofValidationContext
...
         {
-            Logger.LogDebug("Error parsing DPoP token: {error}", ex.Message);
-            result.SetError("Malformed DPoP token.");
-            return Task.CompletedTask;
+            Logger.LogDebug("Failed to parse DPoP cnf claim: {JsonExceptionMessage}", e.Message);
+            result.SetError("Invalid 'cnf' value.");
         }
+    }
+
+    internal bool ValidateJwk(DPoPProofValidationContext context, DPoPProofValidationResult result)
```
bhazen, line 161: It doesn't look like we use this return type and this is inconsistent with the other validation methods. Do we need it?

aspnetcore-authentication-jwtbearer/test/AspNetCore.Authentication.JwtBearer.Tests/DPoP/PayloadTests.cs
```diff
@@ -9,78 +9,80 @@ public class PayloadTests : DPoPProofValidatorTestBase
...
+    public Task missing_payload_fails()
     {
         Result.Payload = null;
 
-        await ProofValidator.ValidatePayload(Context, Result);
+        ProofValidator.ValidatePayload(Context, Result);
 
         Result.ShouldBeInvalidProofWithDescription("Missing payload");
         ProofValidator.ReplayCacheShouldNotBeCalled();
+        return Task.CompletedTask;
```
bhazen, line 20: Should we change the return type of this test and the test below it in this file as has been done with other tests in this file?

bhazen (changes requested):

aspnetcore-authentication-jwtbearer/src/AspNetCore.Authentication.JwtBearer/DPoP/DPoPProofValidator.cs
josephdecock: No, we don't need it. Switched to void.

aspnetcore-authentication-jwtbearer/test/AspNetCore.Authentication.JwtBearer.Tests/DPoP/PayloadTests.cs
josephdecock: Yes, thanks.

bhazen (approved):

PR #1998, opened 2025-05-06
Ensure WWW-Authenticate uses a single HTTP header 

khalidabuhakmeh:
Introduce tests for ProtectedResourceErrorResult to ensure WWW-Authenticate header formatting is correct. Refactor header construction values into a single string.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+# Visual Studio Code workspace options
+.vscode/settings.json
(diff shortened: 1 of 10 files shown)


josephdecock:
I've updated the tests to use Shoudly so that this fix will merge more easily into main. I've also made some minor tooling fixes to the release branches. They now all increment minver patch versions, and this branch was rebased on to that change. I also ran into a vscode settings file not getting ignored, so I added that to gitignore.

bhazen (approved):

PR #2002, opened 2025-05-07
Remove "Duende.IdentityServer.Experimental" service name from telemetry

maartenba:
As documented at https://docs.duendesoftware.com/identityserver/diagnostics/otel/, the detailed metrics moved from the "Duende.IdentityServer.Experimental" service name to the stable "Duende.IdentityServer". These metrics are instrumented by the IdentityServer middleware and services, and track usage of specific flows and features.

If you have subscribed to the associated meters, for example in an OpenTelemetry dashboard, you'll need to update the service name going forward.

Associated docs PR: https://github.com/DuendeSoftware/docs.duendesoftware.com/pull/713

### DIFF
--- a/identity-server/src/Telemetry/Telemetry.cs
+++ b/identity-server/src/Telemetry/Telemetry.cs
@@ -23,7 +23,8 @@ public static class Telemetry
     /// <summary>
     /// Service name used for the experimental non stable counters from Duende IdentityServer
     /// </summary>
-    public const string ServiceNameExperimental = "Duende.IdentityServer.Experimental";
+    [Obsolete("The experimental service name will be removed in a future version. The associated meter is now using the Duende.IdentityServer service name.")]
+    public const string ServiceNameExperimental = ServiceName;
 
     /// <summary>
     /// Metrics configuration.
@@ -93,7 +94,8 @@ public static class Telemetry
         /// <summary>
         /// Meter for experimental counters from IdentityServer
         /// </summary>
-        public static readonly Meter ExperimentalMeter = new Meter(ServiceNameExperimental, ServiceVersion);
+        [Obsolete("The ExperimentalMeter will be removed in a future version. Use the default Meter instead.")]
+        public static readonly Meter ExperimentalMeter = Meter;
 
         /// <summary>
         /// Counter for active requests.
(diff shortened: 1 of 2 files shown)


identity-server/src/Telemetry/Telemetry.cs
```diff
@@ -23,7 +23,8 @@ public static class Telemetry
     /// <summary>
     /// Service name used for the experimental non stable counters from Duende IdentityServer
     /// </summary>
-    public const string ServiceNameExperimental = "Duende.IdentityServer.Experimental";
+    [Obsolete("The experimental service name will be removed in a future version. The associated meter is now using the Duende.IdentityServer service name.")]
```
josephdecock, line 26: I think we can just delete this entirely. If someone is using the old service constant, the fact that we're not using it anymore means that probably whatever they're doing with it is not semantically correct, even if it compiles with warnings. IMO a compilation failure in that case is better. Do you have a use case in mind where someone would want this constant, even though we don't have any counters built with it?

identity-server/src/Telemetry/Telemetry.cs
```diff
@@ -93,7 +94,8 @@ public static class TagValues
         /// <summary>
         /// Meter for experimental counters from IdentityServer
         /// </summary>
-        public static readonly Meter ExperimentalMeter = new Meter(ServiceNameExperimental, ServiceVersion);
+        [Obsolete("The ExperimentalMeter will be removed in a future version. Use the default Meter instead.")]
+        public static readonly Meter ExperimentalMeter = Meter;
```
josephdecock, line 98: Similarly to `ServiceNameExperimental`, let's just get rid of this Meter (unless you think there's a use case where folks would create their own counters using the ExperimentalMeter?)

identity-server/src/Telemetry/Telemetry.cs
maartenba: Will drop it, agree with the suggestion. Thanks Joe!

identity-server/test/IdentityServer.UnitTests/Telemetry/TelemetryTests.cs
```diff
@@ -0,0 +1,78 @@
...
+        List<(string Name, long Value, KeyValuePair<string,object?>[] Tags)> measurements = new();
+        
+        listener.SetMeasurementEventCallback<long>((instrument, measurement, tags, _) =>
+        {
+            measurements.Add((instrument.Name, measurement, tags.ToArray()));
+        });
+        
+        listener.InstrumentPublished = (instrument, meterListener) =>
+        {
+            if (instrument.Meter.Name == "Duende.IdentityServer")
```
maartenba, line 70: Can this use the constant reference from `Telemetry`,  or would that defeat the purpose of checking for the string literal value?

josephdecock (approved):

maartenba:
@josephdecock can I merge?

damianh:
> @josephdecock can I merge?

Let the IS / PO team handle the merges pls.

josephdecock:
Done!

PR #2006, opened 2025-05-09
Options of Allowed Signing Algorithms JWTs and DPoP Proof Tokens

bhazen:
Adds new options for allowed signing algorithms for JWTs and DPoP proof tokens for scenarios when the supported algorithms need to be restricted such as the FAPI 2.0 Profile.

### DIFF
--- a/identity-server/src/IdentityServer/Configuration/DependencyInjection/Options/IdentityServerOptions.cs
+++ b/identity-server/src/IdentityServer/Configuration/DependencyInjection/Options/IdentityServerOptions.cs
@@ -217,7 +217,7 @@ public class IdentityServerOptions
     public TimeSpan JwtValidationClockSkew { get; set; } = TimeSpan.FromSeconds(10);
 
     /// <summary>
-    /// The allowed algorithms for JWT validation. All JWTs validated use this
+    /// The allowed algorithms for JWT validation. Except for DPoP proofs, all JWTs validated by IdentityServer use this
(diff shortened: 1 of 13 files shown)


identity-server/src/IdentityServer/Configuration/DependencyInjection/Options/IdentityServerOptions.cs
```diff
@@ -215,6 +215,17 @@ public class IdentityServerOptions
     /// </summary>
     public TimeSpan JwtValidationClockSkew { get; set; } = TimeSpan.FromSeconds(10);
 
+    /// <summary>
+    /// The allowed algorithms for JWT validation. All JWTs validated use this
```
josephdecock, line 220:
```suggestion
    /// The allowed algorithms for JWT validation. Except for DPoP proofs, all JWTs validated by IdentityServer use this
```

josephdecock (commented):
Please add tests for the dpop proof case, and consider adding tests for the discovery response as well (I think there are existing integration tests so it should be straightfoward).

josephdecock:
> > Please add tests for the dpop proof case, and consider adding tests for the discovery response as well (I think there are existing integration tests so it should be straightfoward).
> 
> @josephdecock There's an existing test in DPoPProofValidationTests with a `invalid_alg_should_fail_validation` test which covers this behavior. Were there other scenarios you wanted covered?

Oh, nice. I guess we could make use of the new option in that test though? That would cover what we will do in the FAPI 2.0 case.

PR #2007, opened 2025-05-09
Add and test unhandled exception logging filter logic

khalidabuhakmeh:

The UnhandledExceptionLoggingFilter is current a `Func` which can either be a single invocation or a multi-cast delegate. This can lead to unintened behaviors if a developer were to use a `+=` instead of `=`.

This pull request suggests testing to see if there is an invocation list on the delegate and then determining the result of all filters by checking each one. If all are `true` then the exception should be logged/emitted.


josephdecock:
This seems reasonable to me, but @bhazen would you please review in detail?

bhazen (approved):

PR #2013, opened 2025-05-11
Fix tabs -> spaces in several csproj's

damianh:
ocd strikes again


josephdecock:
Maybe we should make editorconfig deal with this for us?

josephdecock (approved):

Erwinvandervalk (approved):

damianh:
> Maybe we should make editorconfig deal with this for us?

It does. I think these pre-dated this. `dotnet format` doesn't look at `.csproj` formatting.

PR #2023, opened 2025-05-20
Registered Implementations Diagnostic Entry

bhazen:
Stacks on #2022 to add a diagnostic entry for implementations registered for our interfaces

### DIFF
--- a/identity-server/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/Core.cs
+++ b/identity-server/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/Core.cs
@@ -214,6 +214,7 @@ public static class IdentityServerBuilderExtensionsCore
 
         builder.Services.AddSingleton<IDiagnosticEntry, AssemblyInfoDiagnosticEntry>();
         builder.Services.AddSingleton<IDiagnosticEntry, AuthSchemeInfoDiagnosticEntry>();
+        builder.Services.AddSingleton<IDiagnosticEntry, RegisteredImplementationsDiagnosticEntry>();
         builder.Services.AddSingleton<DiagnosticSummary>();
         builder.Services.AddHostedService<DiagnosticHostedService>();
 
--- a/identity-server/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/Core.cs
+++ b/identity-server/src/IdentityServer/Configuration/DependencyInjection/BuilderExtensions/Core.cs
@@ -214,6 +214,7 @@ public static class IdentityServerBuilderExtensionsCore
 
         builder.Services.AddSingleton<IDiagnosticEntry, AssemblyInfoDiagnosticEntry>();
         builder.Services.AddSingleton<IDiagnosticEntry, AuthSchemeInfoDiagnosticEntry>();
+        builder.Services.AddSingleton(new ServiceCollectionAccessor(builder.Services));
         builder.Services.AddSingleton<IDiagnosticEntry, RegisteredImplementationsDiagnosticEntry>();
         builder.Services.AddSingleton<DiagnosticSummary>();
         builder.Services.AddHostedService<DiagnosticHostedService>();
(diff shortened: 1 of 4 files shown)


josephdecock (commented):
1. How are we going to handle services that can validly have multiple implementations, such as ISecretParser and IExtensionGrantValidator? 
2. How do we decide which public interfaces are included in this list?
3. If we add new interfaces in the future, do we have to remember to update this?

bhazen:
> 1. How are we going to handle services that can validly have multiple implementations, such as ISecretParser and IExtensionGrantValidator?
> 2. How do we decide which public interfaces are included in this list?
> 3. If we add new interfaces in the future, do we have to remember to update this?

@josephdecock to answer those questions:
1. We'll need to update to handle those differently. I will make that change now.
2. I started with all the public interfaces for services and stores as a starting point. I'm open go suggestions on what criteria we should use.
3. As it stands now yes. We could look at doing something with reflection in the constructor to build up the list of interfaces if we wanted something more automated. We'd need to add a way to indicate when an interface can have multiple implementations.

josephdecock:
>     1. We'll need to update to handle those differently. I will make that change now. We could call GetServices for every interface and have an array of implementations for everything. Do you have thoughts on that vs only having an array when we expect multiple implementations?

Always an array seems simplest, and if multiple implementations are registered unexpectedly we would want to know.

>     2. I started with all the public interfaces for services and stores as a starting point. I'm open go suggestions on what criteria we should use.

I think most public interfaces that will always be accessible from the diagnostics host should be captured. 

- All of the stores should be included. Right now some are missing. I would organize by structure a bit more. You have stores from D.IS.Storage.Stores and D.IS.Stores all in one list (side note, we have a bit of mess in our namespaces, don't we?). Concretely:
   - Add IBackChannelAuthenticationRequestStore and IRefreshTokenStore
   - Organize IAuthorizationParametersMessageStore, IConsentMessageStore, IMessageStore, IServerSideTicketStore, ISigningCredentialStore, and IValidationKeysStore into their own section, since they are in a different nuget package
- Include all of the Validators
- Include all of the ResponseGenerators
- There are quite a few things that are almost never (or at least rarely) replaced, but it would be very interesting in a support case if they were. I guess I would include these:
   - ICancellationTokenProvider 
   - ICorsPolicyService
   - IPersistentGrantSerializer
   - IClock
   - IConcurrencyLock<T> (I think there are multiple implementations for different types T)
  
I know it's a lot. Sorry.
 
We probably have to leave out the interfaces from the other packages, like D.IS.Configuration and D.IS.EntityFramework.Storage.

>     3. As it stands now yes. We could look at doing something with reflection in the constructor to build up the list of interfaces if we wanted something more automated. We'd potentially need to add a way to indicate when an interface can have multiple implementations depending on what direction we decide for point number 1.

Could we do reflection in a test? I did that in the mappings to make sure that we didn't forget a model property in an entity (or vice versa).

bhazen:
> > Do you have thoughts on that vs only having an array when we expect multiple implementations?
> 
> Always an array seems simplest, and if multiple implementations are registered unexpectedly we would want to know.
> 
> > I started with all the public interfaces for services and stores as a starting point. I'm open go suggestions on what criteria we should use.
> 
> I think most public interfaces that will always be accessible from the diagnostics host should be captured.
> 
> * All of the stores should be included. Right now some are missing. I would organize by structure a bit more. You have stores from D.IS.Storage.Stores and D.IS.Stores all in one list (side note, we have a bit of mess in our namespaces, don't we?). Concretely:
>   
>   * Add IBackChannelAuthenticationRequestStore and IRefreshTokenStore
>   * Organize IAuthorizationParametersMessageStore, IConsentMessageStore, IMessageStore, IServerSideTicketStore, ISigningCredentialStore, and IValidationKeysStore into their own section, since they are in a different nuget package
> * Include all of the Validators
> * Include all of the ResponseGenerators
> * There are quite a few things that are almost never (or at least rarely) replaced, but it would be very interesting in a support case if they were. I guess I would include these:
>   
>   * ICancellationTokenProvider
>   * ICorsPolicyService
>   * IPersistentGrantSerializer
>   * IClock
>   * IConcurrencyLock (I think there are multiple implementations for different types T)
> 
> I know it's a lot. Sorry.
> 
> We probably have to leave out the interfaces from the other packages, like D.IS.Configuration and D.IS.EntityFramework.Storage.
> 
> > We could look at doing something with reflection in the constructor to build up the list of interfaces if we wanted something more automated. We'd potentially need to add a way to indicate when an interface can have multiple implementations depending on what direction we decide for point number 1.
> 
> Could we do reflection in a test? I did that in the mappings to make sure that we didn't forget a model property in an entity (or vice versa).

I've made these changes. Per our off-line discussion about the open generic types: I've included those manually for now and added a comment documenting why that choice was made.

josephdecock (approved):

PR #2026, opened 2025-05-22
Reworked supported algorithm options

josephdecock:
- Removed AllowedJwtAlgorithms
- Added SupportedClientAssertionSigningAlgorithms, which controls allowed algs for private_key_jwt
- Added SupportedRequestObjectSigningAlgorithms, which controls allowed algs for JAR
- No alg validation is now done in TokenValidator, instead we rely on the fact that we are validating tokens signed by IdentityServer, and that its signing keys (and their algorithms) are configurable.

### DIFF
--- a/identity-server/src/IdentityServer/ResponseHandling/Default/DiscoveryResponseGenerator.cs
+++ b/identity-server/src/IdentityServer/ResponseHandling/Default/DiscoveryResponseGenerator.cs
@@ -328,7 +328,8 @@ public class DiscoveryResponseGenerator : IDiscoveryResponseGenerator
             }
             entries.Add(OidcConstants.Discovery.TokenEndpointAuthenticationMethodsSupported, types);
 
-            if (types.Contains(OidcConstants.EndpointAuthenticationMethods.PrivateKeyJwt))
+            if (types.Contains(OidcConstants.EndpointAuthenticationMethods.PrivateKeyJwt) &&
+                !IEnumerableExtensions.IsNullOrEmpty(Options.SupportedClientAssertionSigningAlgorithms))
             {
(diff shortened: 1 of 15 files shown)


identity-server/src/IdentityServer/ResponseHandling/Default/DiscoveryResponseGenerator.cs
```diff
@@ -326,8 +326,13 @@ where scope.ShowInDiscoveryDocument
                 types.Add(OidcConstants.EndpointAuthenticationMethods.TlsClientAuth);
                 types.Add(OidcConstants.EndpointAuthenticationMethods.SelfSignedTlsClientAuth);
             }
-
             entries.Add(OidcConstants.Discovery.TokenEndpointAuthenticationMethodsSupported, types);
+
+            if (types.Contains(OidcConstants.EndpointAuthenticationMethods.PrivateKeyJwt))
+            {
+                entries.Add(OidcConstants.Discovery.TokenEndpointAuthSigningAlgorithmsSupported,
```
bhazen, line 333: Do we need checks on any of these to ensure the list is not empty? Is adding an empty list to the document valid? I recall having a discussion around at least in some cases empty lists being invalid, but don't recall if that was for very specific scenarios or a general thing.

josephdecock: Yes, that was actually my intention and how I wrote the xmldoc. I believed that our json serialization for disco was omitting null/empty collections, but I went and added tests to prove it. Turns out I was wrong about disco serialization - the tests failed! So I've now updated the response generator to omit each of the 3 options when they are null or empty.

bhazen (approved):

PR #2035, opened 2025-06-02
Fix Parameterized Scope Client

bhazen:
* Adds parameterized scope client to IdentityServer solution filter
* Updates parameterized scope client to work with Aspire config

### DIFF
--- a/identity-server/aspire/AppHost/AppHost.csproj
+++ b/identity-server/aspire/AppHost/AppHost.csproj
@@ -17,56 +17,46 @@
   </ItemGroup>
 
   <ItemGroup>
-
-    <!-- Database Projects -->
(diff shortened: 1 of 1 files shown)


damianh (approved):

identity-server/aspire/AppHost/AppHost.csproj
```diff
@@ -19,6 +19,7 @@
   <ItemGroup>
 
     <!-- Database Projects -->
+    <ProjectReference Include="..\..\clients\src\ConsoleParameterizedScopeClient\ConsoleParameterizedScopeClient.csproj" />
```
josephdecock, line 22: Can we please either remove the grouping of project references and alphabetize so that we can add projects via ide without caring about where in the file they end up, or hand edit this line down to the section where it belongs?

bhazen: Removed the grouping and sorted to match how things display in the solution tree

josephdecock (approved):

PR #2060, opened 2025-06-13
Apply V4 style in in-memory host and fix logout issue

Erwinvandervalk:
Applies v4 style config to BFF. 

found an issue with logout & cookie configuration when using BFF V4 style without a frontend. Updated tests to cover this scenario and also fixed the issue.

I found some issues that could only be explained by having 3 different wireup mechanisms. So I also applied the test matrix to much more tests. 


bff/src/Bff.Yarp/Internal/ProxyDataExtensionLoader.cs
damianh, line 1: What's going on here?

bff/src/Bff/Configuration/IDataExtensionLoader.cs
damianh, line 1: Also here?

bff/src/Bff.Yarp/Internal/ProxyDataExtensionLoader.cs
Erwinvandervalk: I think they were somehow still open in my editor and it 'restored' it.

damianh (approved):

PR #2072, opened 2025-06-19
Add support for multiple frontends to the SQL session storage. 

Erwinvandervalk:
Sessions are now partitioned 'per frontend'.

To verify this is true, the tests are now run against a sql lite db, not against in-memory (which doesn't check unique constraints)


bff/src/Bff.EntityFramework/Internal/UserSessionStore.cs
```diff
@@ -3,19 +3,20 @@
...
 /// Entity framework core implementation of IUserSessionStore
 /// </summary>
-internal class UserSessionStore(IOptions<DataProtectionOptions> options, ISessionDbContext sessionDbContext, ILogger<UserSessionStore> logger) : IUserSessionStore, IUserSessionStoreCleanup
+internal class UserSessionStore(
+    IUserSessionPartitionKeyBuilder partitionKeyBuilder,
+    ISessionDbContext sessionDbContext,
+    ILogger<UserSessionStore> logger) : IUserSessionStore, IUserSessionStoreCleanup
 {
-    private readonly string? _applicationDiscriminator = options.Value.ApplicationDiscriminator;
+    private string? partitionKey => partitionKeyBuilder.BuildPartitionKey();
```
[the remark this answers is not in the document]
Erwinvandervalk: The in-memory implementation is different from the sql implementation. The in-memory doesn't support null discriminators. SQL can support it (depending on the selected database). 

Unfortunately, some databases don't include null values in indexes, so the default implementation will not use a null. But implementors can decide to use a null value.

bff/migrations/Directory.Build.props
```diff
@@ -3,7 +3,6 @@
   <PropertyGroup>
     <RootNamespace>$(AssemblyName)</RootNamespace>
     <AssemblyName>Duende.$(MSBuildProjectName)</AssemblyName>
-    <TargetFramework>net9.0</TargetFramework>
   </PropertyGroup>
   <Import Project="../../samples.props" />
 </Project>
```
damianh, line 8: Add blank last line

Erwinvandervalk: done

damianh (approved):

PR #2080, opened 2025-06-27
mTLS with DPoP

josephdecock:
If both an mTLS certificate and dpop proof are sent to the token endpoint, bind the resulting token to the proof. We assume mtls is being used for client auth (which is the only reasonable way both mTLS and DPOP would be sent).


identity-server/test/IdentityServer.IntegrationTests/Common/IdentityServerPipeline.cs
```diff
@@ -47,6 +49,8 @@ public class IdentityServerPipeline
     public const string CheckSessionEndpoint = BaseUrl + "/connect/checksession";
     public const string ParEndpoint = BaseUrl + "/connect/par";
 
+    // TODO - add mtls aliases
```
bhazen, line 52: Was this TODO something meant to be part of this PR?

bhazen (approved):

identity-server/test/IdentityServer.IntegrationTests/Common/IdentityServerPipeline.cs
josephdecock: Yes, thanks!

PR #2099, opened 2025-07-07
Address CA1307 Violations

bhazen:
**What issue does this PR address?**
Stacks on #2098 to remove [CA1307](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1307) from the NoWarn list and address the resulting build warnings.

**Important: Any code or remarks in your Pull Request are under the following terms:**

If You provide us with any comments, bug reports, feedback, enhancements, or modifications proposed or suggested by You for the Software, such Feedback is provided on a non-confidential basis (notwithstanding any notice to the contrary You may include in any accompanying communication), and Licensor shall have the right to use such Feedback at its discretion, including, but not limited to the incorporation of such suggested changes into the Software. You hereby grant Licensor a perpetual, irrevocable, transferable, sublicensable, nonexclusive license under all rights necessary to incorporate and use your Feedback for any purpose, including to make and sell any products and services.

(see [our license](https://duendesoftware.com/license/identityserver.pdf), section 7)

PR #2100, opened 2025-07-07
Address CA1308 Violations

bhazen:
Stacks on #2099 to remove [CA1308](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1308) from the NoWarn list and address the resulting build warnings.


identity-server/src/EntityFramework/Services/CorsPolicyService.cs
```diff
@@ -52,7 +52,9 @@ public CorsPolicyService(IConfigurationDbContext dbContext, ILogger<CorsPolicySe
     /// <returns></returns>
     public async Task<bool> IsOriginAllowedAsync(string origin)
     {
+#pragma warning disable CA1308 // this has historically been normalized to lower case
```
josephdecock, line 55: I was initially concerned, because CA1308's guidance says to not make security decisions based on strings that you have called ToLowerInvariant on, but in this case it is fine and in fact what RFC 3986 says we should do. We're comparing origins (scheme, host and port), and it says:

> the scheme and host are case-insensitive and therefore should be normalized to lowercase.

josephdecock (approved):
Wow, CA1308

PR #2112, opened 2025-07-16
Address CA1001 Violations

bhazen:
Stacks on #2102 to remove [CA1001](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1001) from the NoWarn list and address the resulting build warnings.


identity-server/src/IdentityServer/Infrastructure/ConcurrencyLock/DefaultConcurrencyLock.cs
```diff
@@ -24,4 +25,25 @@ public Task<bool> LockAsync(int millisecondsTimeout)
 
     /// <inheritdoc/>
     public void Unlock() => Lock.Release();
+
+    public void Dispose()
+    {
+        Dispose(true);
+        GC.SuppressFinalize(this);
```
josephdecock, line 32: I'm curious what the intent here is, since we don't have a finalizer in this class.

bhazen: Line 31 here addresses [CA1063](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1063) and line 32 addresses [CA1816](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1816) which were new violations after implementing `IDisposable` in this class

josephdecock: Oh, that's really interesting. CA1816 specifically mentions unsealed classes. So I guess this is required because a derived type could declare a finalizer. That would be a hell of a messy situation. But it's okay in the end - the docs say that SuppressFinalize doesn't do anything if there is no finalizer.

PR #2116, opened 2025-07-17
Address CA1040 Violations

bhazen:
Stacks on #2115 to remove [CA1040](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1040) from the NoWarn list and address the resulting build warnings.


bhazen:
> My (weak) opinion is that we should add this to the list of intentionally suppressed rules, since the only times we do it are intentional. What do you think?

The one question I have before doing that: would we ever want to be alerted to when we do this in the future? I don't know that it much matters since this rule is more stylistic than doing this will cause problems.

josephdecock:
I think I just disagree with the rule.

> you are probably using an interface as a marker or a way to identify a group of types. If this identification will occur at run time, the correct way to accomplish this is to use a custom attribute. Use the presence or absence of the attribute, or the properties of the attribute, to identify the target types.

If I recall correctly, we use these marker interfaces to resolve the marked types from DI, and attributes wouldn't be a replacement.

bhazen:
Looking at the usages, only `IServerSideSessionMarksMarker` is ever registered with DI. The only interfaces look more to creating a shared return type. Changing any of these now though would be a breaking change that wouldn't be appropriate for a non-major release.

I'm fine with keeping the rule on the NoWarn list and documenting it if that's what you'd like to do.

josephdecock:
Ok, yes, I'd like to do that.

josephdecock:
Thanks!

josephdecock (approved):

PR #2126, opened 2025-07-23
Callback Option for Dynamic Providers

bhazen:
Adds a new option for dynamic providers to allow using a callback to detect the auth scheme over the existing path option. The new option will only be used if it has been assigned a value, otherwise the existing path prefix option will be used as it is today.


identity-server/src/IdentityServer/Configuration/DependencyInjection/Options/DynamicProviderOptions.cs
```diff
@@ -22,6 +22,12 @@ public class DynamicProviderOptions
     /// </summary>
     public PathString PathPrefix { get; set; } = "/federation";
 
+    /// <summary>
+    /// If set, this callback will be used over <see cref="PathPrefix"/> to determine if the request is for a dynamic provider.
+    /// If the request is a match, the callback should return the name of the scheme which matched.
+    /// </summary>
+    public Func<HttpContext, string?>? PathMatchingCallback { get; set; }
```
josephdecock, line 29: Should this be async? Do we think users would want to interact with a data store in the callback?

josephdecock (approved):

PR #2166, opened 2025-08-07
Log diagnostics to filesystem in Development builds only

josephdecock:


identity-server/templates/src/IdentityServerAspNetIdentity/HostingExtensions.cs
```diff
@@ -1,15 +1,54 @@
...
+                    formatProvider: CultureInfo.InvariantCulture);
+                if (builder.Environment.IsDevelopment())
+                {
+                    consoleLogger.Filter.ByExcluding(Matching.FromSource("Duende.IdentityServer.Diagnostics.Summary"));
+                }
+            });
+            if (builder.Environment.IsDevelopment())
+            {
+                Console.WriteLine(builder.Environment.EnvironmentName);
+                Console.WriteLine("Hi!");
```
khalidabuhakmeh, line 34: Not sure you meant to keep this "Hi!" here.

bhazen: @josephdecock is this leftover from debugging? We have these same two `Console.Writeline` statements in the `IdentityServer` template as well.

bhazen (approved):

PR #2183, opened 2025-08-22
Set synchronization delay to 0

josephdecock:
In all hosts, set the InitializationSynchronizationDelay to 0. This
makes it faster for the hosts to reach a healthy state, and is safe to
do as they are never run in a load balanced environment.

### DIFF
--- a/identity-server/hosts/AspNetIdentity/IdentityServerExtensions.cs
+++ b/identity-server/hosts/AspNetIdentity/IdentityServerExtensions.cs
@@ -10,7 +10,16 @@ internal static class IdentityServerExtensions
 {
     internal static WebApplicationBuilder ConfigureIdentityServer(this WebApplicationBuilder builder)
     {
-        builder.Services.AddIdentityServer()
+        builder.Services.AddIdentityServer(opt =>
+            {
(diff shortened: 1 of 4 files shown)


identity-server/hosts/AspNetIdentity/IdentityServerExtensions.cs
```diff
@@ -10,7 +10,9 @@ internal static class IdentityServerExtensions
 {
     internal static WebApplicationBuilder ConfigureIdentityServer(this WebApplicationBuilder builder)
     {
-        builder.Services.AddIdentityServer()
+        builder.Services.AddIdentityServer(opt =>
+            opt.KeyManagement.InitializationSynchronizationDelay = TimeSpan.Zero // For fast startup
```
bhazen, line 14: I think we may want to be more verbose in the comment we're using when setting this option. If someone is looking at our source and only sees "For fast startup" and doesn't read the docs to understand the purpose of this setting it could cause issues for that person. Perhaps we expand to include something such as "this is typically not desirable for a production scenario" or include what's in the PR description about it being safe since this won't run in a load balanced scenario. What do you think?

bhazen (approved):

PR #2189, opened 2025-08-28
RFC 8414 support

bhazen:
Adds support for [RFC 8414: OAuth 2.0 Authorization Server Metadata](https://datatracker.ietf.org/doc/html/rfc8414). 

This also required adjustments to the endpoint routing to accommodate a path that isn't an exact match to support issue identifiers with path components.

### DIFF
--- a/identity-server/test/IdentityServer.UnitTests/Licensing/v2/DiagnosticEntries/EndpointUsageDiagnosticEntryTests.cs
+++ b/identity-server/test/IdentityServer.UnitTests/Licensing/v2/DiagnosticEntries/EndpointUsageDiagnosticEntryTests.cs
@@ -41,7 +41,7 @@ public class EndpointUsageDiagnosticEntryTests
         var endpointUsage = result.RootElement.GetProperty("EndpointUsage");
         foreach (var endpoint in _endpoints)
         {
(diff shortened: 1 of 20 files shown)


identity-server/test/IdentityServer.UnitTests/Licensing/v2/DiagnosticEntries/EndpointUsageDiagnosticEntryTests.cs
```diff
@@ -70,7 +70,7 @@ public async Task Should_Handle_Unknown_Endpoints()
         endpointUsage.GetProperty("other").GetInt64().ShouldBe(1);
         foreach (var endpoint in _endpoints)
         {
-            endpointUsage.GetProperty(endpoint).GetInt64().ShouldBe(0);
+            endpointUsage.GetProperty(endpoint.Value!).GetInt64().ShouldBe(0);
```
josephdecock, line 73: Can we please add a test that verifies that /.well-known/oauth-authorization-server/bogus-unknown-issuer-subpath is counted as an unknown call, and not a disco call?

josephdecock: Oh, I guess I was wrong, let's not do anything here then.

PR #2229, opened 2025-09-29
Catch potential OperationCanceledException in DiagnosticHostedService

wcabus:

Some [testing frameworks fail](https://github.com/orgs/DuendeSoftware/discussions/348) when running integration tests on the new `DiagnosticHostedService`: when the service stops, the `PeriodicTimer` can throw an `OperationCanceledException`.

I wasn't able to simulate this in a unit test when calling the normal `StartAsync` and `StopAsync` hosted service methods, which is why I added an internal test-only method.


wcabus:
Imo, this fix doesn't need to be backported to 7.3, I'll leave the final decision with you both.

